PR #20, opened 2025-12-29
Dashboard <> Deposit address

Megha-Dev-19:
image
image

### DIFF
--- a/nt-be/src/handlers/token/metadata.rs
+++ b/nt-be/src/handlers/token/metadata.rs
@@ -60,9 +60,11 @@ struct RefSdkToken {
     pub price: Option<f64>,
     #[serde(rename = "priceUpdatedAt")]
     pub price_updated_at: Option<String>,
+    #[serde(rename = "price_updated_at")]
     pub price_updated_at_snake_case: Option<String>,
     #[serde(rename = "chainName")]
     pub chain_name: Option<String>,
+    #[serde(rename = "chain_name")]
     pub chain_name_snake_case: Option<String>,
     pub error: Option<String>,
 }


nt-fe/app/(treasury)/[treasuryId]/dashboard/components/deposit-modal.tsx
```diff
@@ -0,0 +1,477 @@
...
+        </div>
+
+        {/* Content */}
+        <div className="px-4 pt-4 pb-6 space-y-2">
+          <p className="text-sm font-semibold">
+            Select asset and network to see deposit address
+          </p>
+
+          {/* Error Message */}
+          {error && (
```
Comment (akorchyn) [inline nt-fe/app/(treasury)/[treasuryId]/dashboard/components/deposit-modal.tsx]: Should be done by react-hook-form rather than you own handling

akorchyn (changes requested):
Sorry, I'm not ready to merge this. Please use components, and do not create your own implementation for things that could be used from Shadcn or already implemented.

E.g. modal.tsx

We don't want to have 10 implementations of the same things

nt-fe/components/ui/alert.tsx
```diff
@@ -11,7 +11,9 @@ const alertVariants = cva(
         default: "bg-card text-card-foreground",
         info: "bg-general-info-background-faded text-general-info-foreground border border-general-info-border",
         destructive:
-          "text-destructive bg-card [&>svg]:text-current *:data-[slot=alert-description]:text-destructive/90",
+          "bg-red-50 dark:bg-red-950/50 border-red-200 dark:border-red-800 text-red-800 dark:text-red-200 [&>svg]:text-red-600 dark:[&>svg]:text-red-400",
+        warning:
+          "bg-yellow-50 dark:bg-yellow-950/50 border-yellow-200 dark:border-yellow-800 text-yellow-800 dark:text-yellow-200 [&>svg]:text-yellow-600 dark:[&>svg]:text-yellow-400",
```
Megha-Dev-19: Our UI has different error and warning state, so I have added these colors, do you think we should stay with library one or our own implementation? @akorchyn

akorchyn: Ideally, we should create a wrapper around the original alert with an override similar to @/components/button, overrides components/ui/button.

We don't want to really change the default implementation, as we might, from time to time, update that component, and it will override our settings.

nt-be/src/handlers/token/metadata.rs
```diff
@@ -43,17 +44,59 @@ pub struct TokenMetadata {
...
     pub price_updated_at: Option<String>,
-    #[serde(rename = "chainName")]
-    pub chain_name: String,
+    pub chain_name: Option<String>,
+    pub error: Option<String>,
+}
+
+/// Manual parser for RefSdkToken to handle duplicate keys in API response
+/// The REF SDK API returns both snake_case and camelCase versions of some fields
+fn parse_ref_sdk_token(value: &Value) -> Option<RefSdkToken> {
```
akorchyn: Don't really like this...

akorchyn: Let's just fetch as before and then iterate and clean up rather than this

PR #24, opened 2026-01-03
fix: handle day overflow in monthly interval increment

petersalomonsen:
## Problem

The monthly interval increment function would panic when incrementing from dates like January 31st to February, because February doesn't have 31 days.

## Solution

Use chrono's `checked_add_months` method which automatically handles day overflow by clamping to the last valid day of the target month.

## Changes

- Updated `Interval::increment()` to use `checked_add_months(Months::new(1))`
- Added comprehensive unit tests covering edge cases:
  - Jan 31 → Feb 29 (leap year)
  - Jan 31 → Feb 28 (non-leap year)  
  - Mar 31 → Apr 30
  - May 31 → Jun 30
  - Jan 29-30 → Feb (both leap and non-leap years)

## Testing

All 13 unit tests pass:
```
cargo test interval_increment --lib
```

### DIFF
--- a/nt-be/src/handlers/balance_changes/history.rs
+++ b/nt-be/src/handlers/balance_changes/history.rs
@@ -42,9 +42,7 @@ impl Interval {
             Interval::Hourly => datetime + chrono::Duration::hours(1),
             Interval::Daily => datetime + chrono::Duration::days(1),
             Interval::Weekly => datetime + chrono::Duration::weeks(1),
-            Interval::Monthly => datetime
-                .checked_add_months(Months::new(1))
-                .expect("Failed to increment month"),
+            Interval::Monthly => datetime.checked_add_months(Months::new(1)).unwrap(),
         }
     }
 }


nt-be/src/handlers/balance_changes/history.rs
```diff
@@ -32,31 +32,19 @@ impl Interval {
...
-
-                // Keep the same day, time, and timezone
-                datetime
-                    .with_year(new_year)
-                    .and_then(|dt| dt.with_month(new_month))
-                    .expect("Failed to increment month")
-            }
+            Interval::Monthly => datetime
+                .checked_add_months(Months::new(1))
+                .expect("Failed to increment month"),
```
akorchyn: When does it crash ? 

Ideally we should not use unwrap and expect over the code 

Is it safe to assume that it never happens ?

petersalomonsen: yes, we can assume it will never crash. changed to `unwrap` now

akorchyn: I would rather unwrap_or_default()

but I probably will iterate on it later and fight all the unwraps/expects in the codebase

PR #95, opened 2026-01-20
CSS fixes for members and deposit modal

Megha-Dev-19:


nt-fe/components/button-with-tooltip.tsx
```diff
@@ -0,0 +1,43 @@
+import { Button } from "@/components/button";
+import {
+  Tooltip,
+  TooltipContent,
+  TooltipTrigger,
+} from "@/components/ui/tooltip";
```
akorchyn: Use components/tooltip

As we use another design for that

PR #104, opened 2026-01-22
fix: recent activity with staking

akorchyn:


nt-be/src/routes/balance_changes.rs
```diff
@@ -156,6 +156,7 @@ pub async fn get_recent_activity(
         WHERE account_id = $1
           AND counterparty != 'SNAPSHOT'
           AND counterparty != 'NOT_REGISTERED'
+          AND counterparty != 'STAKING_SNAPSHOT'
```
petersalomonsen: For staking snapshot I would keep it, since it will be a balance change to the previous staking snapshot entry

PR #142, opened 2026-02-04
feat: prioritize balance change history sync on UI interaction

### LINKED ISSUE #135: Prioritize updating treasury balance change history on UI interaction
## Problem

When a user executes a transaction through the treasury UI, they expect to see it reflected immediately. Currently, the backend monitoring cycle processes all accounts sequentially by staleness — so the user's treasury may be queued behind other accounts, causing a noticeable delay before the transaction appears in their history.

## Solution

Introduce a priority sync mechanism so that accounts actively used in the UI get immediate attention, **in addition to** the normal monitoring cycle.

### How it works

- A new `dirty_at` timestamp field on `monitored_accounts` marks accounts that need priority attention
- When the UI triggers a transaction, it calls a new API endpoint that sets `dirty_at` (defaulting to 24 hours back)
- A dedicated background watcher polls for dirty accounts every 5 seconds and spawns parallel tasks to fill recent gaps
- Each dirty task fills gaps in **reverse block height order** (most recent first), so the latest transactions appear in the UI first
- Once all gaps in the `dirty_at → now` window are filled, `dirty_at` is conditionally cleared
- Staking reward tracking is excluded from dirty tasks to keep them fast — it continues in the normal monitoring cycle

### Key properties

- **Additive** — the existing monitor cycle is unchanged; dirty accounts get double attention
- **One task per account** — no duplicate work if the same account is re-dirtied while a task is running
- **Race-safe** — conditional clear ensures a new dirty flag set mid-task isn't lost

See [implementation plan](docs/implementation-plans/dirty-account-priority-monitoring.md) for full design details.

petersalomonsen:
## Summary

Implements priority sync for monitored accounts so that treasuries actively used in the UI get their balance change history updated immediately, rather than waiting for the normal sequential monitoring cycle.

Closes #135

### Dirty account priority monitoring (core feature)
- Added `dirty_at` timestamp column to `monitored_accounts` via migration
- Implemented `dirty_monitor` background watcher that polls for dirty accounts every 5 seconds and spawns parallel gap-filling tasks in reverse block height order
- Added API endpoint to mark accounts as dirty (sets `dirty_at` to 24h back by default)
- Race-safe conditional clear ensures new dirty flags set mid-task aren't lost
- One task per account prevents duplicate work

**Key files:**
- `nt-be/src/handlers/balance_changes/dirty_monitor.rs` — core dirty monitor logic
- `nt-be/migrations/20260204000003_add_dirty_at_to_monitored_accounts.sql`
- `nt-be/src/routes/monitored_accounts.rs` — API route for marking dirty
- `nt-be/tests/dirty_monitor_e2e_test.rs` — E2E test for the feature

### Subscription system
- Added subscription plans configuration, API handlers, and frontend hooks
- Database migrations for subscription tables
- Frontend subscription API integration

### Transport error retries
- Added `with_transport_retry` helper (3 retries, exponential backoff) for transient RPC errors (`DispatchGone`, `SendError`, connection timeouts)
- Applied to all RPC call sites in `block_info.rs` and `tx_resolver.rs`

### Test & E2E fixes for camelCase API migration
- Updated `monitored_accounts_test.rs` payloads/assertions to camelCase
- Updated bulk payment E2E test with NEP-413 auth flow and camelCase payloads
- Reduced `NUM_RECIPIENTS` from 100 to 25 to match API limit

### Sandbox configuration
- Added missing `BULK_PAYMENT_SIGNER` env var to `supervisord.conf`
- Fixed sandbox FullAccess key and contract deserialization

### Frontend updates
- Ledger wallet: derivation path, UI confirmations/errors, auto-open NEAR app
- Members page mobile responsiveness
- Bulk payment UI improvements
- Sign-in component updates

### Documentation
- Added `docs/AI_SUBSCRIPTION_GUIDE.md` and `docs/PRICING.md`
- Added implementation plan for dirty account priority monitoring

## Test plan

- [x] `cargo check` compiles
- [x] `cargo clippy` passes
- [x] Backend tests pass (including dirty monitor E2E test)
- [x] Bulk payment E2E test passes (with 25 recipients)
- [x] Dirty monitor E2E verifies block heights and receipt IDs

### DIFF
--- a/e2e-tests/bulk-payment/dao-bulk-payment-flow.js
+++ b/e2e-tests/bulk-payment/dao-bulk-payment-flow.js
@@ -719,10 +719,10 @@ assert.match(listId, /^[0-9a-f]{64}$/, 'list_id must be hex-encoded');
 // Step 7b: Verify API rejects submission with WRONG hash (payload doesn't match list_id)
 console.log('\n🔒 Testing API rejection with mismatched hash...');
 const wrongHashResponse = await apiRequest('/api/bulk-payment/submit-list', 'POST', {
-  list_id: listId,
-  submitter_id: daoAccountId,
-  dao_contract_id: daoAccountId,
-  token_id: 'native',
+  listId: listId,
+  submitterId: daoAccountId,
+  daoContractId: daoAccountId,
+  tokenId: 'native',
   // Tamper with payments - change first recipient's amount
   payments: payments.map((p, i) => i === 0 ? { ...p, amount: '999' } : p),
 }, true); // expectError = true
@@ -735,10 +735,10 @@ console.log(`✅ API correctly rejected tampered payload: ${wrongHashResponse.er
 // Step 7c: Verify API rejects submission WITHOUT a DAO proposal
 console.log('\n🔒 Testing API rejection without DAO proposal...');
 const rejectResponse = await apiRequest('/api/bulk-payment/submit-list', 'POST', {
-  list_id: listId,
-  submitter_id: daoAccountId,
-  dao_contract_id: daoAccountId,
-  token_id: 'native',
+  listId: listId,
+  submitterId: daoAccountId,
+  daoContractId: daoAccountId,
+  tokenId: 'native',
   payments,
 }, true); // expectError = true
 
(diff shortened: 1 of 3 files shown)


akorchyn (approved):

akorchyn:
I think it would be sufficient with async approach. But I don't mind threaded version too

PR #164, opened 2026-02-09
Nearn integration in members page

Megha-Dev-19:
- Only allow AddProposal role for nearn account
- Added query param `?member=nearn-io.near&roles=requestor` which can be used on NEARN to add member, @akorchyn you don't need to pass `roles` since we already have checks in frontend for `nearn-io.near` account
- Updated the validation, moved all roles validation to edit screen from the review request page like in design

### DIFF
--- a/nt-fe/app/(treasury)/[treasuryId]/members/page.tsx
+++ b/nt-fe/app/(treasury)/[treasuryId]/members/page.tsx
@@ -958,69 +958,61 @@ export default function MembersPage() {
                                 </TableCell>
                                 <TableCell className="pr-6">
                                     <div className="flex justify-end gap-2 md:opacity-0 group-hover:opacity-100 transition-opacity">
-                                        <Tooltip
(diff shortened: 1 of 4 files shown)


nt-fe/app/(treasury)/[treasuryId]/members/page.tsx
```diff
@@ -798,39 +958,69 @@ export default function MembersPage() {
...
-                                            variant="ghost"
-                                            size="icon"
-                                            onClick={() => {
-                                                setMemberToDelete(member);
-                                                setIsDeleteModalOpen(true);
-                                            }}
-                                            disabled={hasPendingMemberRequest}
-                                            className="h-8 w-8 text-destructive hover:text-destructive hover:bg-destructive/10"
+                                            <span>
+                                                <AuthButton
```
Comment (akorchyn) [inline nt-fe/app/(treasury)/[treasuryId]/members/page.tsx]: I believe AuthButton has a tooltip parameter if not it's  a good idea to forward it into Button itself

PR #195, opened 2026-02-15
feat: completeness API, tx hash resolution, swap detection fixes

### LINKED ISSUE #156: Add real-time balance history completeness check
## Problem

There's no way for the UI or exports to know whether balance history for a DAO is complete. This affects:
- **Exports:** Users may export partial data without realizing it
- **Recent activity:** After a new on-chain transaction, there's a sync lag before it appears in the DB
- **Historical data:** We may not have backfilled all the way to account creation

## Why Not a Stored Flag?

A stored `history_complete` boolean doesn't work because completeness is constantly changing:
- New on-chain transactions create a forward gap until the sync catches up
- Gap filling progressively resolves interior and backward gaps
- A flag would go stale the moment a new transaction happens

Instead, compute completeness on-demand. The existing `find_gaps` query is cheap (index-backed LAG window function, milliseconds for typical accounts) and always accurate.

## How Completeness Is Determined

No new columns or migrations needed. Everything can be derived from existing data:

1. **Interior gaps** — run existing `find_gaps()` per token. If `balance_before[i] != balance_after[i-1]`, there's a gap.
2. **Backward completeness** — depends on the token type:
   - **NEAR:** Named accounts (we don't track implicit accounts) are always created with a nonzero balance to cover storage. So if the earliest NEAR record has `balance_before = 0`, we've reached the account creation. If `balance_before > 0`, we **definitely** haven't — there must be an earlier creation transaction.
   - **FT tokens (NEP-141):** `balance_before = 0` is **not** reliable — an account could sell all tokens (balance → 0), then receive more later. Instead, check whether the account was registered with the FT contract (`storage_balance_of`) at a block before the earliest record. If not registered, there's no earlier history. This requires one archival RPC call per FT token.
   - **Intents tokens (NEP-245 on intents.near):** No storage registration, but the same backward walk works. The gap filler progressively walks back ~7 days per cycle via SNAPSHOTs (see `fill_gap_to_past`, lookback of 600,000 blocks). When the backward search reaches a block where the balance is 0, we've found the beginning. A SNAPSHOT with `balance_before > 0` as the earliest record means the backward walk is still in progress. Note: history cannot go further back than the creation of the intents.near contract itself.
3. **Forward freshness** — expose `last_synced_at` (already tracked on `monitored_accounts`) so the consumer can judge how stale the data might be.

---

### Task 1: Add Completeness API Endpoint

**`GET /api/balance-history/completeness?accountId=X`**

Computes and returns real-time completeness:

```json
{
  "account_id": "treasury.near",
  "last_synced_at": "2025-01-15T12:00:00Z",
  "tokens": [
    {
      "token_id": "near",
      "has_gaps": false,
      "gap_count": 0,
      "reaches_beginning": true
    },
    {
      "token_id": "usdt.tether-token.near",
      "has_gaps": true,
      "gap_count": 2,
      "reaches_beginning": false
    },
    {
      "token_id": "intents.near:nep141:eth.omft.near",
      "has_gaps": false,
      "gap_count": 0,
      "reaches_beginning": false
    }
  ]
}
```

**Implementation:**
For each token tracked by the account:
1. Run `find_gaps(pool, account_id, token_id, current_block)` — existing function, returns interior gaps
2. Check backward completeness (token-type dependent):
   - **NEAR:** Query earliest record's `balance_before` — if 0, history is complete back to creation
   - **FT tokens:** Query earliest record's `balance_before`. If 0, additionally call `storage_balance_of` at a block before the earliest record (archival RPC) to confirm the account wasn't registered earlier. If not registered → complete. If registered → we may be missing earlier history
   - **Intents tokens:** Same as NEAR/FT — earliest record's `balance_before = 0` (and not a SNAPSHOT) means we've reached the beginning. A SNAPSHOT with `balance_before > 0` means the backward walk is still in progress

### Task 2: Include `last_synced_at` in Balance History Responses

Add `last_synced_at` to the response metadata for `/api/balance-changes`, `/api/recent-activity`, and `/api/balance-history/chart` so the UI can show a "data as of" indicator without a separate API call.

### Task 3: Tests

- Completeness endpoint returns correct `gap_count` (leverages existing `gap_detector` tests)
- NEAR: `reaches_beginning: true` when earliest `balance_before = 0`
- NEAR: `reaches_beginning: false` when earliest `balance_before > 0`
- FT: `reaches_beginning` correctly reflects `storage_balance_of` check
- Intents: `reaches_beginning` based on backward SNAPSHOT walk (same logic as NEAR/FT)
- `last_synced_at` appears in balance history responses

---

## Files Summary

| File | Changes |
|------|---------|
| `nt-be/src/routes/balance_changes.rs` | Add completeness endpoint, include `last_synced_at` in responses |
| `nt-be/src/handlers/balance_changes/gap_detector.rs` | Possibly add a helper that combines gap check + earliest balance check |

No new migrations. No new columns.

## Acceptance Criteria

- [ ] Completeness endpoint returns real-time gap info per token
- [ ] NEAR: `reaches_beginning` is true only when earliest `balance_before = 0`
- [ ] FT: `reaches_beginning` uses archival `storage_balance_of` check
- [ ] Intents: `reaches_beginning` based on backward SNAPSHOT walk
- [ ] `last_synced_at` is included in balance history API responses
- [ ] `cargo build` and `cargo test` pass

petersalomonsen:
Closes #156

### Completeness API
- Add `GET /api/balance-history/completeness` endpoint with `from`/`to` time-range parameters
- Returns per-token gap details (start/end block, block times, surrounding balances) instead of just a count
- Add `find_gaps_in_time_range` to gap detector for time-bounded gap queries
- Include `last_synced_at` in balance-changes, recent-activity, and chart API responses

### Transaction hash resolution (`tx_resolver`)
- Refactor `resolve_receipt_to_transaction` to use `EXPERIMENTAL_receipt` RPC — no longer needs `predecessor_id` as input
- Handle **Data receipts**: find the actual Action receipt via `account_changes` on the receiver, then trace its signer
- Handle **meta-transactions** (DELEGATE_ACTION): `EXPERIMENTAL_receipt` returns the real on-chain signer (e.g. `treasury-factory.near`), not the user
- Extract `find_transaction_on_signer` and `find_action_receipt_at_block` helpers

### Missing tx hash backfill
- Add `resolve_missing_tx_hashes` in gap_filler — runs each dirty/account monitor cycle (limit 10 per cycle)
- Uses OFFSET-based pagination to avoid retrying the same failed records within a cycle
- Wired into both `account_monitor` and `dirty_monitor`

### Balance query & binary search fixes
- Remove silent block-fallback loops from `ft.rs`, `near.rs`, `intents.rs` — balance queries now return errors on `UnknownBlock`
- Binary search handles `UnknownBlock` directly by skipping non-existing blocks (`left = mid + 1; continue`)
- This fixes the root cause of zero-amount UNKNOWN records (binary search converging on wrong blocks)

### Swap detection fix
- Fix `detected_swaps` unique constraint: change from `UNIQUE(account_id, fulfillment_receipt_id)` to `UNIQUE(account_id, solver_transaction_hash)`
- The old constraint caused all intents swaps to collide on empty string (intents balance changes have no receipt_id)
- Make `fulfillment_receipt_id` nullable
- Migration: `20260216000001_fix_detected_swaps_unique_constraint.sql`

image

## Test plan
- [x] `cargo clippy` clean
- [x] `cargo test --lib` — 163 tests pass
- [x] New unit tests for Data receipt resolution (`test_resolve_meta_transaction_receipt`, `test_resolve_data_receipt_wrap_near`)
- [x] Integration tests for completeness API and swap detection
- [x] Verified swap detection finds 5 swaps (was 1) for test account after fix

### DIFF
--- a/nt-be/src/handlers/balance_changes/utils.rs
+++ b/nt-be/src/handlers/balance_changes/utils.rs
@@ -6,14 +6,10 @@ use sqlx::types::chrono::{DateTime, Utc};
 use std::future::Future;
 use tokio::time::{Duration, sleep};
 
-const MAX_TRANSPORT_RETRIES: u32 = 5;
+const MAX_TRANSPORT_RETRIES: u32 = 3;
 
 /// Check if an error is a transient transport/network error that should be retried
 pub fn is_transport_error(err_debug: &str) -> bool {
-    // Don't retry 422 — these are permanent "block unavailable" errors
-    if err_debug.contains("422") {
-        return false;
-    }
     err_debug.contains("TransportError")
         || err_debug.contains("SendError")
         || err_debug.contains("DispatchGone")
@@ -21,16 +17,12 @@ pub fn is_transport_error(err_debug: &str) -> bool {
         || err_debug.contains("error sending request")
         || err_debug.contains("connection")
         || err_debug.contains("timed out")
-        || err_debug.contains("429")
-        || err_debug.contains("Too Many Requests")
-        || err_debug.contains("LIMIT_EXCEEDED")
 }
 
-/// Retry an async operation on transient errors with exponential backoff.
+/// Retry an async operation on transient transport errors with exponential backoff.
 ///
-/// Retries up to 5 times with delays of 500ms, 1s, 2s, 4s, 8s.
-/// Covers transport errors (connection, timeout) and rate limiting (429).
-/// Non-transient errors are returned immediately without retrying.
+/// Retries up to 3 times with delays of 200ms, 400ms, 800ms.
+/// Non-transport errors are returned immediately without retrying.
 pub async fn with_transport_retry<T, E, F, Fut>(label: &str, mut make_call: F) -> Result<T, E>
 where
     F: FnMut() -> Fut,
@@ -39,9 +31,9 @@ where
 {
     for attempt in 0..=MAX_TRANSPORT_RETRIES {
         if attempt > 0 {
-            let delay_ms = 500 * 2u64.pow(attempt - 1);
+            let delay_ms = 200 * 2u64.pow(attempt - 1);
             log::warn!(
-                "{}: transient error, retrying in {}ms (attempt {}/{})",
+                "{}: transport error, retrying in {}ms (attempt {}/{})",
                 label,
                 delay_ms,
                 attempt + 1,
--- a/nt-be/tests/intents_tokens_metadata_test.rs
+++ b/nt-be/tests/intents_tokens_metadata_test.rs
@@ -1,0 +1,134 @@
+//! Integration test for intents token metadata discovery
+//!
+//! Tests that running the monitor cycle discovers and stores metadata for intents tokens
+//! by extracting the actual FT contract and querying it.
+
+mod common;
+
+use nt_be::handlers::balance_changes::account_monitor::run_monitor_cycle;
+
+/// Test that intents token metadata is discovered during monitoring
+#[tokio::test]
+async fn test_intents_tokens_metadata_discovery() {
+    common::load_test_env();
+
+    // Get test database URL
+    let db_url =
+        std::env::var("DATABASE_URL").expect("DATABASE_URL must be set for integration tests");
+
+    // Connect to database
+    let pool = sqlx::postgres::PgPoolOptions::new()
+        .max_connections(5)
+        .connect(&db_url)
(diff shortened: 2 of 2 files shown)


petersalomonsen:
@Megha-Dev-19 
image

PR #225, opened 2026-02-18
fix: persist bulk payment worker queue to database

### LINKED ISSUE #224: fix: bulk payment worker loses pending lists on restart and called payout_batch on non-approved lists
## Bug

The bulk payment payout worker has two issues:

### 1. Calling payout_batch on non-approved lists (fixed in #223)

The worker was calling `payout_batch` for every list in its queue without first checking the on-chain approval status. Lists still in `Pending` status caused the contract to panic with "List must be Approved to process payments", resulting in failed transactions that cost gas.

**Fix (PR #223):** Added a `view_list` call before `payout_batch` to check status. Only approved lists proceed.

### 2. In-memory queue loses pending lists on restart (this issue)

The worker tracks pending payment lists in a `lazy_static Mutex<HashSet<String>>`. When the server restarts, all pending lists are lost from memory and never get their `payout_batch` called — payments stall silently with no recovery mechanism.

## Root Cause

`PENDING_LISTS` in `worker.rs` only lives in process memory. There is no persistence layer and no startup recovery.

## Proposed Fix

1. Replace the in-memory `HashSet` with a `pending_payment_lists` PostgreSQL table
2. The worker SELECTs from this table on each poll cycle, so lists present at startup are automatically processed
3. Lists are DELETEd when completed, rejected, or not found on-chain
4. Add a verification step to the e2e test (`dao-bulk-payment-flow.js`) that confirms the worker does NOT call `payout_batch` while a list is still pending approval

## Changes

- New migration: `pending_payment_lists` table
- `worker.rs`: Replace `lazy_static` HashSet with sqlx queries
- `submit.rs`: Pass `&PgPool` to `add_pending_list`
- Remove `lazy_static` crate dependency
- Extend e2e test with payout_batch regression check

petersalomonsen:
## Summary

- Replace in-memory `HashSet` with a `pending_payment_lists` PostgreSQL table so the worker queue survives server restarts
- Remove `lazy_static` dependency (only used for the in-memory set)
- Add e2e regression test (Step 9b in `dao-bulk-payment-flow.js`) that scans sandbox blocks to verify the worker does **not** call `payout_batch` while a list is still pending approval
- Add `.dockerignore` to reduce Docker build context from 16GB to ~12MB
- Use pre-built `cargo-near` binary in Dockerfile (faster builds)
- Upgrade `near-sandbox` to 0.3.6 for ARM64/aarch64 support in sandbox-init
- Set default `NUM_RECIPIENTS` to 25 (matching API limit)

Closes #224

## Test plan

- [x] Verified regression test catches the bug (commented out approval check → test exits with `payout_batch` calls detected)
- [x] Verified test passes with approval check enabled (0 `payout_batch` calls during 117 blocks while list was pending)
- [x] `cargo clippy -- -D warnings` passes
- [x] `cargo sqlx prepare --check` passes
- [x] CI Docker build succeeds
- [x] `npm run test:docker` — e2e test passes including Step 9b verification

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,7 @@
+**/target
+**/node_modules
+.git
+nt-fe
+e2e-tests
+docs
+.github
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ node_modules
 .claude
 .env
 nt-fe/public/ledger-wallet/*.js
+sandbox/sandbox-init/target
--- a/e2e-tests/bulk-payment/dao-bulk-payment-flow.js
+++ b/e2e-tests/bulk-payment/dao-bulk-payment-flow.js
@@ -55,8 +55,8 @@ const CONFIG = {
   BULK_PAYMENT_CONTRACT_ID: process.env.BULK_PAYMENT_CONTRACT_ID || 'bulk-payment.near',
   
   // Test parameters
-  // Note: 250 is the max before gas limits are exceeded for list deserialization (~156 TGas for storage read)
-  NUM_RECIPIENTS: parseInt(process.env.NUM_RECIPIENTS || '250', 10),
+  // Note: 25 is the max per bulk payment API limit
+  NUM_RECIPIENTS: parseInt(process.env.NUM_RECIPIENTS || '25', 10),
   PAYMENT_AMOUNT: process.env.PAYMENT_AMOUNT || '100000000000000000000000', // 0.1 NEAR per recipient
   
   // Genesis account credentials (default sandbox genesis account from near-sandbox-rs - PUBLIC TEST KEY)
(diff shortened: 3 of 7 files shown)


petersalomonsen:
@race-of-sloths include

nt-be/migrations/20260218000001_create_pending_payment_lists.sql
```diff
@@ -0,0 +1,4 @@
+CREATE TABLE pending_payment_lists (
```
akorchyn: Let's add dao for stats

nt-be/src/handlers/bulkpayment/worker.rs
```diff
@@ -52,34 +51,52 @@ impl PaymentListStatus {
...
 }
 
-/// Add a list_id to the set of pending lists to be processed by the worker
-pub async fn add_pending_list(list_id: String) {
-    log::info!("Adding list {} to payout worker queue", list_id);
-    let mut pending = PENDING_LISTS.lock().await;
-    pending.insert(list_id);
-    log::info!("Payout worker queue now has {} lists", pending.len());
+/// Remove a list_id from the pending payment lists table.
+async fn remove_pending_list(pool: &PgPool, list_id: &str) {
```
akorchyn: I think we should still keep it for stats

PR #238, opened 2026-02-19
fix: filter out tiny NEAR amounts from recent activity

### LINKED ISSUE #236: Filter out tiny NEAR amounts (< 0.01) from recent activity
## Quick Fix

Filter out NEAR balance changes smaller than 0.01 NEAR from the recent activity API. These are gas costs or storage deposits that clutter the activity feed and are not meaningful to users.

## Example

A DAO proposal transferring 100 NEAR to `sponsor.trezu.near` also shows entries like `-0.0006 NEAR` or `-0.0999 NEAR` in the activity feed. These should be hidden.

## Suggested Implementation

Add `exclude_near_dust: bool` to `BalanceChangeFilters` in `query_builder.rs`. In `build_where_conditions()`, when true, add:

```sql
NOT (token_id = 'near' AND ABS(amount) < 0.01)
```

No bind parameter needed — it's a hardcoded threshold. Set `exclude_near_dust: true` only in the `get_recent_activity` handler (not in the general `/api/balance-changes` endpoint, which is used for accounting/export).

The filter and count query in `get_recent_activity` both use `BalanceChangeFilters`, and the data query goes through `BalanceChangesQuery` → `get_balance_changes_internal`. The `BalanceChangesQuery` struct would also need the field so it can be forwarded to `BalanceChangeFilters` inside `get_balance_changes_internal`.

## Related

Quick fix for #235.

petersalomonsen:
image

## Summary
- Filter out NEAR balance changes < 0.01 from the recent activity API (gas/storage noise)
- Balance-changes API (accounting/export) is unaffected — dust records remain visible
- Added integration test verifying both endpoints behave correctly

## Test plan
- [x] Unit tests pass (`cargo test query_builder`)
- [x] Integration test `recent_activity_dust_filter_test` verifies dust is excluded from recent activity but preserved in balance-changes API
- [x] Manually verified on dev DB that balance-changes endpoint still returns all records

Closes #236

### DIFF
--- a/e2e-tests/bulk-payment/dao-bulk-payment-flow.js
+++ b/e2e-tests/bulk-payment/dao-bulk-payment-flow.js
@@ -104,19 +104,20 @@ function generateImplicitAccountId(index) {
  * 1. Sorts object keys alphabetically
  * 2. Sorts payments by recipient
  */
-function generateListId(submitterId, tokenId, payments) {
+function generateListId(submitterId, tokenId, payments, timestamp) {
   // Sort payments by recipient for deterministic ordering (must match API)
   const sortedPayments = [...payments].sort((a, b) => a.recipient.localeCompare(b.recipient));
-  
+
   // Create canonical JSON with alphabetically sorted keys (matches Rust serde_json)
-  // Key order: payments, submitter, token_id (alphabetical)
+  // Key order: payments, submitter, timestamp, token_id (alphabetical)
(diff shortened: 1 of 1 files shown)


race-of-sloths:
🥁  __Score it!__

@petersalomonsen, please ask your peers to score the PR using `@race-of-sloths score [1/2/3/5/8/13]`.
If no scoring is provided within 24 hours, this PR will be scored as 1 🦥

PR #307, opened 2026-03-02
Improvments

Megha-Dev-19:
- Dialogs/modals now open as drawers on smaller screens
- Deposit network is no longer auto-selected when multiple networks are available
- Added relative timestamps in Recent Activity
- Fixed layout issues on Preferences and Members pages for smaller screens
- Added network-specific links for the receiver address
- Improved bulk import: more descriptive errors and more flexible parsing
- Fixed primary button text color when customized by the user

### DIFF
--- a/nt-fe/components/ui/dialog.tsx
+++ b/nt-fe/components/ui/dialog.tsx
@@ -60,19 +60,7 @@ function DialogContent({
       <DialogPrimitive.Content
         data-slot="dialog-content"
         className={cn(
-          // Base styles
-          "bg-background fixed z-50 grid gap-4 border p-6 shadow-lg duration-200",
-          // Animation base
-          "data-[state=open]:animate-in data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=open]:fade-in-0",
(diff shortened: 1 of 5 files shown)


nt-fe/components/ui/dialog.tsx
```diff
@@ -60,7 +60,18 @@ function DialogContent({
       <DialogPrimitive.Content
```
akorchyn: I would prefer make overrides on level of ours components

components/modal.tsx

nt-fe/lib/blockchain-utils.ts
```diff
@@ -138,3 +138,80 @@ export function requiresCrossChainValidation(chainName?: string, residency?: str
...
+        case "aptos":
+            return `https://aptoscan.com/account/${address}`;
+
+        case "cardano":
+            return `https://cardanoscan.io/address/${address}`;
+
+        case "unknown":
+        default:
+            // Return empty string for unknown chains - no link will be shown
+            return "";
```
akorchyn: lets better use null as more explicit

PR #320, opened 2026-03-04
Adds create-treasury cli

elliotBraem:
Simple script to create treasury via near-cli-rs

Screenshot 2026-03-03 at 9 27 31 PM

Success: https://trezu.app/chicago.sputnik-dao.near


akorchyn:
Hello Elliot. Not sure if we have the capacity to maintain this CLI tool at the current stage of things. 
I think you could move it as part of near-cli-rs  (as it already supports voting with sputnik-dao), but less trezu specific

Or separate into some devhub/near repo, but I doubt it would be actively maintained :(

PR #399, opened 2026-03-23
Added owned and other assets distinction in deposit modal

Megha-Dev-19:
image
image


Megha-Dev-19:
Yeah, I agree. While working on it, I did consider making it reusable across both, but there are some differences in usability between them, so I avoided it at the time to not over-engineer.

Since you’ve also pointed it out, I’ll go ahead and create a ticket to refactor it into a reusable function.

PR #429, opened 2026-03-30
ft: NEAR and FT Lockups

Megha-Dev-19:
- Added available, locked, earning tab in assets and chart
image

image

- Added scheduler to claim unlocked tokens, runs every 6 hours 

Note: I have tested with multiple ft lockups, wrap.near, usdt -> viewing and claiming
NEAR lockup (with full/partial staked balance)

### DIFF
--- a/nt-be/.sqlx/query-09167e480bc3123777af345c030d8fac0aed9f4de50c9c7d3e358bd667003e4e.json
+++ b/nt-be/.sqlx/query-09167e480bc3123777af345c030d8fac0aed9f4de50c9c7d3e358bd667003e4e.json
@@ -1,0 +1,40 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "\n        SELECT dao_account_id, instance_id, token_account_id, ft_registered_at\n        FROM ft_lockup_dao_schedules\n        WHERE (next_claim_at IS NULL OR next_claim_at <= NOW())\n        ORDER BY COALESCE(next_claim_at, NOW()) ASC\n        LIMIT $1\n        ",
+  "describe": {
+    "columns": [
+      {
+        "ordinal": 0,
+        "name": "dao_account_id",
+        "type_info": "Text"
+      },
+      {
+        "ordinal": 1,
+        "name": "instance_id",
+        "type_info": "Text"
+      },
+      {
+        "ordinal": 2,
+        "name": "token_account_id",
+        "type_info": "Text"
+      },
+      {
+        "ordinal": 3,
+        "name": "ft_registered_at",
+        "type_info": "Timestamptz"
+      }
+    ],
+    "parameters": {
+      "Left": [
+        "Int8"
+      ]
+    },
+    "nullable": [
+      false,
+      false,
+      false,
+      true
+    ]
+  },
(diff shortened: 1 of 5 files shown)


nt-be/migrations/20260330000001_create_ft_lockup_dao_schedules.sql
```diff
@@ -0,0 +1,17 @@
+CREATE TABLE IF NOT EXISTS ft_lockup_dao_schedules (
+    dao_account_id TEXT NOT NULL,
+    instance_id TEXT NOT NULL,
+    token_account_id TEXT NOT NULL,
+    session_interval_seconds BIGINT,
+    start_timestamp_seconds BIGINT,
+    is_ft_registered BOOLEAN NOT NULL DEFAULT FALSE,
```
akorchyn: It's nice to use ft_registered_at for bool props; this way, we can still have a boolean flow. (VALUE IS NULL OR NOT)
But also provides some debug info

nt-fe/components/auth-button.tsx
```diff
@@ -46,6 +46,8 @@ function ErrorMessage({
     return (
         <Button
             {...props}
+            disabled
+            onClick={undefined}
             tooltipContent={message}
```
akorchyn: Does the error tooltip work on mobile/pc?

nt-fe/lib/dashboard-balance-view.ts
```diff
@@ -0,0 +1,146 @@
...
+        availableRaw = totalRaw.sub(lockedRaw);
+    } else if (token.balance.type === "Staked") {
+        const staked = token.balance.staking.stakedBalance;
+        const unstaked = token.balance.staking.unstakedBalance;
+        totalRaw = staked.add(unstaked);
+        earningRaw = staked;
+    } else if (token.balance.type === "Vested") {
+        const lockup = token.balance.lockup;
+        const staked = lockup.staked;
+        const nonStakedLocked = lockup.unvested.sub(staked);
```
akorchyn: Should we have here availableRaw too? As it might be available for near lockups

nt-fe/lib/dashboard-balance-view.ts
```diff
@@ -0,0 +1,146 @@
...
+    let earningRaw = ZERO;
+
+    if (token.balance.type === "Standard") {
+        totalRaw = token.balance.total;
+        lockedRaw = token.balance.locked.gt(totalRaw)
+            ? totalRaw
+            : token.balance.locked;
+        availableRaw = totalRaw.sub(lockedRaw);
+    } else if (token.balance.type === "Staked") {
+        const staked = token.balance.staking.stakedBalance;
```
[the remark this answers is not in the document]
Megha-Dev-19: hmmm, I think it will become too complicated, because they will have to be withdrawn first to use it, so I think it makes sense to be under 'Available for withdrawal' inside earning tab
image

nt-fe/components/auth-button.tsx
akorchyn: We also show an error message when hovering on PC and as on click on the mobile. So I was wondering if it propagates

PR #491, opened 2026-04-14
Onboarding ui fixes

Megha-Dev-19:
- Added animation on landing page
- Fixed small screen UI
- Back from create treasury takes back to last questions instead of 1
- Fixed near icon display

### DIFF
--- a/nt-fe/app/(init)/page.tsx
+++ b/nt-fe/app/(init)/page.tsx
@@ -438,9 +438,23 @@ export function Content() {
     }
 
     return (
-        <div className="h-screen w-full overflow-hidden bg-muted p-4 md:px-8 md:py-6">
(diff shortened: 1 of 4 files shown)


nt-fe/app/(init)/page.tsx
```diff
@@ -438,9 +438,23 @@ export function Content() {
     }
 
     return (
-        <div className="h-screen w-full overflow-hidden bg-muted p-4 md:px-8 md:py-6">
-            <div className="mx-auto flex h-full max-w-[1180px] items-center justify-center">
-                <div className="w-full md:p-6">
+        <div className="min-h-screen w-full overflow-y-auto bg-muted p-4 md:px-8 md:py-6">
```
akorchyn, line 506: I think would be nice to ad more than the only one. Also for typing text and stuff like that. See previous version

akorchyn (approved):

PR #499, opened 2026-04-16
fix: improve Telegram notification reliability and connection UX

Megha-Dev-19:
#### Bugs fixed

- Show only notifications created after the treasury was connected to Telegram get dispatched. Connecting a treasury no longer floods the chat with months of historical proposals/payments.
- Before inserting into `telegram_treasury_connections`, we now call `register_or_refresh_monitored_account` to ensure the DAO exists in `monitored_accounts`. This prevents the `fk_ttc_dao` violation.
- Fixed duplicate proposal notifications by deduplicating `add_proposal` events using transaction hash and preferring the non-zero sibling balance change row.
- Fixed incorrect swap amounts in Telegram messages by using stored human-readable swap amounts directly.

#### UX improvements

- Connect now behaves as “sync selected treasuries” for the chat (unselected treasuries are removed).
- Connect success message now clearly shows selected and removed counts.
- Disconnect now sends clear status updates:
  - if some treasuries remain, the chat is told how many are still connected;
  - if none remain, the bot sends a final reconnect hint and attempts to leave the chat.


akorchyn:
Please take a look also for swap notif, it currently doesn't show decimals correctly. 
Also, sometimes I see several notifications for newly created proposal

nt-be/src/handlers/telegram/connect.rs
```diff
@@ -227,6 +243,25 @@ pub async fn connect_treasuries(
         .map_err(|e| (StatusCode::INTERNAL_SERVER_ERROR, e.to_string()))?;
     }
 
+    // Sync selected treasuries for this chat: remove any previously connected
+    // DAO in this chat that is not part of the current selection.
+    let removed_count = sqlx::query_scalar::<_, i64>(
```
akorchyn, line 248: Let's do it in a transaction. Delete + Update so if any failed we don't commit

akorchyn (approved):

PR #509, opened 2026-04-17
Fix: posthog background capture and ad blocking

Megha-Dev-19:


Megha-Dev-19:
@akorchyn do we want to disable session recording?

akorchyn:
No, not really

nt-fe/instrumentation-client.ts
```diff
@@ -26,5 +26,9 @@ posthog.init(process.env.NEXT_PUBLIC_POSTHOG_PROJECT_TOKEN!, {
     ui_host: "https://us.posthog.com",
     defaults: "2026-01-30",
     capture_exceptions: true,
+    // Keep only explicit capture() calls from app code.
+    autocapture: false,
```
akorchyn: I don't think we should... We want to capture as much as possible

Megha-Dev-19: We do ? I thought Vlad mentioned to reduce the bg activity ? 
Or was it only for surveys?

akorchyn: What Vlad mean that it is a problem that adblocker blocks metrics collection.
So the idea was more about find a way to not get blocked by it rather than reduce

nt-fe/instrumentation-client.ts
```diff
@@ -22,12 +22,11 @@ export const onRouterTransitionStart = Sentry.captureRouterTransitionStart;
 import posthog from "posthog-js";
 
 posthog.init(process.env.NEXT_PUBLIC_POSTHOG_PROJECT_TOKEN!, {
-    api_host: "/ingest",
+    api_host: "/_telemetry",
```
akorchyn, line 25: I think we should use some sneaky name to not get blocked by ad block regex or something

akorchyn (approved):

nt-fe/instrumentation-client.ts
Megha-Dev-19: do you have any suggestions? I tried on few browsers with aggressive block trackers and it worked fine on all, checked with opus/sonnet too and it agrees 😛

PR #520, opened 2026-04-20
Replace SDK fee estimation with Chaindefuser RPC

Megha-Dev-19:


Megha-Dev-19:
Not needed, as intents block certain addresses, wasn't a bug but a feature.

PR #560, opened 2026-04-28
fix: normalize treasury logo values

Megha-Dev-19:


nt-fe/components/treasury-info.tsx
```diff
@@ -1,44 +1,78 @@
...
+    logo: string | null | undefined,
+): string | undefined {
+    if (!logo) return undefined;
+
+    const trimmedLogo = logo.trim();
+    if (!trimmedLogo) return undefined;
+
+    if (/^https?:\/\//i.test(trimmedLogo)) return trimmedLogo;
+
+    return `https://ipfs.near.social/ipfs/${trimmedLogo.replace(/^\/+/, "")}`;
```
akorchyn: i think we should check if it follows hash length or something like that

akorchyn (approved):

PR #562, opened 2026-04-28
fix: apply short expiry only to non-wrap-near exchange flows

Megha-Dev-19:
Short expiry now applies only to true exchange execution flows. Single-action `wrap.near` `near_deposit`/`near_withdraw` requests follow the standard proposal period, preventing them from being marked expired early and still appearing under Pending due to backend/frontend mismatch.

Fixes issue reported by Olha: expired exchange shown in [Pending](https://trezu.app/yurtur-empty-treasury.sputnik-dao.near/requests/16)
image

### DIFF
--- a/nt-be/src/handlers/proposals/scraper.rs
+++ b/nt-be/src/handlers/proposals/scraper.rs
@@ -349,6 +349,41 @@ fn get_current_time_nanos() -> U64 {
     U64::from(nanos as u64)
 }
 
+fn is_short_expiry_exchange(proposal: &Proposal) -> bool {
+    // First, gate short-expiry logic to exchange proposals only.
+    if AssetExchangeInfo::from_proposal(proposal).is_none() {
+        return false;
+    }
+
+    let Some(function_call) = proposal.kind.get("FunctionCall") else {
+        return false;
+    };
+
+    let receiver_id = function_call
+        .get("receiver_id")
+        .and_then(|v| v.as_str())
+        .unwrap_or("");
+    let actions = function_call
+        .get("actions")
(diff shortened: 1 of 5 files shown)


nt-be/src/handlers/proposals/scraper.rs
```diff
@@ -360,13 +395,11 @@ pub fn get_status_display(
...
+            // Exchange proposals use short expiration, except pure wrap/unwrap.
             let expiration_period = if let Some(p) = proposal {
-                if extract_from_description(&p.description, "proposalaction")
-                    == Some("asset-exchange".to_string())
-                {
-                    // 24 hours in nanoseconds
-                    24 * 60 * 60 * 1_000_000_000
+                if is_short_expiry_exchange(p) {
+                    // 1 hour in nanoseconds
+                    60 * 60 * 1_000_000_000
```
akorchyn: why only hour instead of 24?

akorchyn (commented):
Not sure why do we reduce to 1hour?

nt-be/src/handlers/proposals/scraper.rs
akorchyn: It should still be 24, but it should verify:
- if proposal expired -> ignore the exchange, it expired
- If the proposal has not expired, choose the smallest deadline between proposal and custom exchange

Megha-Dev-19: oh I was testing it with 1 hour, completely forgot to change it back.

akorchyn (approved):

PR #567, opened 2026-04-29
feat: add rule-based network sections for guest confidential flows

Megha-Dev-19:
- Added a shared `section-rules` helper to build dropdown sections from simple rules (`title`, `filter`, `disabled`).
- Updated recipient network selection to use caller-provided rules instead of hardcoded section logic.
- Added guest confidential behavior:
  - `Available` shows only `near.com`
  - `For Members Only` shows other networks as disabled
- Applied the same guest section pattern in deposit network selection.
- Improved dropdown row styling in dark mode

### DIFF
--- a/nt-fe/app/(treasury)/[treasuryId]/payments/components/payment-form-section.tsx
+++ b/nt-fe/app/(treasury)/[treasuryId]/payments/components/payment-form-section.tsx
@@ -26,7 +26,12 @@ import { NetworkList } from "@/components/network-list";
 import { Button } from "@/components/button";
 import { UserWithData } from "@/components/user";
 import { FormField } from "@/components/ui/form";
-import { RecipientNetworkSelect } from "./recipient-network-select";
+import { type SectionRule } from "@/lib/section-rules";
+import {
+    NEAR_COM_NETWORK_ID,
+    RecipientNetworkSelect,
+    type RecipientNetworkRuleOption,
+} from "./recipient-network-select";
 import { cn } from "@/lib/utils";
 
 interface PaymentFormSectionProps<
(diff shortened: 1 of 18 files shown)


nt-fe/app/(treasury)/[treasuryId]/payments/components/payment-form-section.tsx
```diff
@@ -126,7 +132,60 @@ export function PaymentFormSection<
         [recipientName, setValue],
     );
 
-    const { isConfidential } = useTreasury();
+    const { isConfidential, isGuestTreasury } = useTreasury();
+
```
Comment (akorchyn) [inline nt-fe/app/(treasury)/[treasuryId]/payments/components/payment-form-section.tsx]: thats kind of weird change. Why do we need to add this for payments as it user anyway wont able to create a payment request

akorchyn (commented):
I'm not sure about payment form to be honest. Can you tell the motivation about this segregation there? 

As for me, it add subtle meaning that user can create simple payment, but it is not the case in fact

nt-fe/app/(treasury)/[treasuryId]/payments/components/payment-form-section.tsx
Comment (Megha-Dev-19) [reply nt-fe/app/(treasury)/[treasuryId]/payments/components/payment-form-section.tsx]: but they can interact with the form that's why I added it.
If you think it doesn't make sense at all, I can remove it, don't have a hard opinion on this.

Comment (akorchyn) [reply nt-fe/app/(treasury)/[treasuryId]/payments/components/payment-form-section.tsx]: I don't like the subtle meaning that something is available to guests. So I would rather disable everything / disabled click on select

Comment (Megha-Dev-19) [reply nt-fe/app/(treasury)/[treasuryId]/payments/components/payment-form-section.tsx]: okay I removed it from the payment form, I talked to Olha and she prefers we have active form so user can see the functionality even when they are not member, which I agree with.

akorchyn (approved):

PR #583, opened 2026-05-04
perf: reduce intents explorer traffic in swap detection and status checks

Megha-Dev-19:


nt-be/src/handlers/balance_changes/account_monitor.rs
```diff
@@ -51,6 +51,34 @@ async fn get_effective_block_floor(
...
+              )
+        ) AS "exists!"
+        "#,
+        account_id
+    )
+    .fetch_one(pool)
+    .await?;
+
+    Ok(has_new)
+}
```
Megha-Dev-19, line 80: @akorchyn what do you think about this, this will reduce the calls quite a lot as we only check if if there is a initiation of swap in `balance_changes`, but do you think there is any possibility we miss some records?

akorchyn: Probably no.

frol: It won't completely help as if there is a simple deposit or payment after the most recent swap, then this function will keep returning true all the time, no?

frol (approved):
I think even though it might not reduce the traffic completely, it is already a good step in the right direction.

nt-be/src/handlers/balance_changes/account_monitor.rs
akorchyn: Yeah, but it might reduce the load. We currently have the issue of rate limits to Explorer api... You can see in render logs

akorchyn: Overall it is also a bit symptomatic fix and all activity should be carefully analyzed and planned ...

PR #608, opened 2026-05-06
fix: align confidential payment request details with public intents metadata

Megha-Dev-19:
- Added confidential payment details to match public flow in request details: `destination network`, `deposit address`, `quote signature`, and `network fee`.
image

- Saved the network fee in the public proposal description at creation time, so request details always show the exact fee used when the proposal was created (because it changes later).
- Simplified destination data to one field: `destinationAssetId` (removed extra name/id duplication).


frol (approved):
It seems that we do too much of conditional logic in our code. I wish we could refactor it into modules with clear separation of concerns instead of marking the decisions inline.

It is not a blocker for the fix, so I am approving it and merging.

PR #635, opened 2026-05-12
fix: normalize NEAR vs near.com network display

Megha-Dev-19:
- Added Staging pill in header.
- In Payment Request details:
  - moved hover info to the Amount
  - always show Destination Network row
- Cleaned casing behavior for near.com. (Near.com -> near.com)
- Fixed network detection edge cases for NEAR token, it showed Native token as Fungible.
- Updated rhea link.
- [x] #636

### DIFF
--- a/nt-fe/components/page-component-layout.tsx
+++ b/nt-fe/components/page-component-layout.tsx
@@ -3,9 +3,10 @@
 import { ArrowLeft, Moon, PanelLeft, Sun } from "lucide-react";
 import { useRouter } from "next/navigation";
 import { useTranslations } from "next-intl";
-import { type ReactNode, useEffect } from "react";
+import { type ReactNode, useEffect, useState } from "react";
 import { Button } from "@/components/button";
 import { LanguageSwitcher } from "@/components/language-switcher";
+import { Pill } from "@/components/pill";
 import { SignIn } from "@/components/sign-in";
 import { SystemStatusBanner } from "@/components/system-status-banner";
 import { ConfidentialBanner } from "@/features/confidential/components/confidential-banner";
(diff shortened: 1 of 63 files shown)


nt-fe/components/page-component-layout.tsx
```diff
@@ -34,13 +35,25 @@ export function PageComponentLayout({
...
     }, [theme]);
 
+    useEffect(() => {
+        if (typeof window !== "undefined") {
+            setShowStagingTag(
+                process.env.NEXT_PUBLIC_STAGING === "true" ||
+                    window.location.hostname.includes("testenv.trezu"),
+            );
+        }
+    }, []);
```
frol: I don't think there is value in this additional check. We already rely on `NEXT_PUBLIC_STAGING` in https://github.com/NEAR-DevHub/trezu/blob/29809f7ed7cffc7ce986ad9a2f4d777717552174/nt-fe/constants/features.ts#L11-L13

Let's remove this useEffect and the useState for `showStatingTag`.

nt-fe/features/proposals/components/amount.tsx
```diff
@@ -18,23 +21,72 @@ interface AmountProps {
...
+
+    const nearTypeLabel = getNearTokenTypeLabel(
+        isNativeNearToken ? "near" : tokenId,
+        resolvedNetwork,
+    );
+
+    if (nearTypeLabel) {
+        return !expandNearComLabel &&
+            nearTypeLabel === "NEAR (near.com) Network"
+            ? "near.com"
```
frol: In general, when we rely on conditions over some text labels, it is a signal of something really odd. Is there really no better way to address it? Why do we need to override the label that is returned by `getNearTokenTypeLabel`? OR, in other words, should `getNearTokenTypeLabel` return `near.com` instead of `NEAR (near.com) Network`?

frol (changes requested):

nt-fe/features/proposals/components/amount.tsx
Megha-Dev-19: yeah completely missed it, fixed!

frol (approved):

Issue #629, opened 2026-05-08
[BUG] Trezu - Confidential Export: Export feature incorrectly accessible in Confidential Treasury via Public Export flow


rknearn-qa:
### Summary

The **Export functionality from Public Treasury** remains accessible when navigating to a **Confidential Treasury**, even though export is expected to be restricted or unavailable in Confidential mode. After opening the Export page in Public Treasury and switching to Confidential Treasury, the Export feature is still visible/available, which violates expected access control rules for Confidential environments.

### Preconditions:

- Public Treasury exists (https://testenv.trezu.app/staging-qa.sputnik-dao.near/dashboard/export)
- Confidential Treasury exists (https://testenv.trezu.app/confidential-staging.sputnik-dao.near)
- User has access to both Public and Confidential Treasury environments

### Steps To Reproduce

1. Open **Public Treasury**
2. Navigate to **Export page**
3. Switch or navigate to **Confidential Treasury**
4. Observe Export availability in Confidential Treasury

### Actual Result:

- Export feature is visible/available in Confidential Treasury
- User can access Export functionality after transitioning from Public Treasury context
- No restriction or “coming soon” state is shown

[confidential-staging.sputnik-dao.near_activity_2025-05-08_2026-05-08.csv](https://github.com/user-attachments/files/27528605/confidential-staging.sputnik-dao.near_activity_2025-05-08_2026-05-08.csv)

<img width="1365" height="694" alt="Image" src="https://github.com/user-attachments/assets/710dcc05-47fc-49b1-9ef5-c2c250fd1a65" />

https://github.com/user-attachments/assets/0a1b86a9-3de6-4afc-9dd2-8bc92feb93d2

### Expected Result:

- Export should **not be available** in Confidential Treasury
- UI should display **“Coming soon”** or equivalent placeholder state
- Access should be fully disabled or hidden for Confidential environments
- No cross-context leakage from Public Treasury functionality

### Special Notes:

```
**Possible Root Cause:**

* State leakage between Public and Confidential Treasury routes (shared session/component state not reset on navigation)
* Missing feature flag separation for Export functionality in Confidential mode
* Route guarding logic not enforcing treasury-type restrictions
* Frontend caching or persisted UI state incorrectly reusing Public Treasury configuration
```

### Workarounds:
None

### Repro Rate:
Always (10/10)

### Setup - OS - Browser - Environment

[...]

Issue #628, opened 2026-05-08
[BUG] Trezu - Confidential Exchange: Estimated time value displayed without unit type


rknearn-qa:
### Summary

In the **Confidential Exchange** request view, the **Estimated time** field is displayed as a raw numeric value (**"10"**) without a corresponding unit label. Unlike the **Public Treasury** implementation (where the value is displayed as **“10 seconds”**), the Confidential Exchange UI does not provide a time unit, leading to unclear and potentially ambiguous interpretation of the value.

### Preconditions:

- Confidential Treasury exists
- Confidential Exchange request is created and accessible (https://testenv.trezu.app/confidential-staging.sputnik-dao.near/requests/41)

### Steps To Reproduce

1. Navigate to **Confidential Exchange** request
2. Open the request details page
3. Locate the **Estimated time** field
4. Observe the displayed value

### Actual Result:

- Estimated time is displayed as a raw number (**“10”**)
- No unit of measurement (e.g., seconds, minutes) is shown

<img width="1365" height="687" alt="Image" src="https://github.com/user-attachments/assets/a9b3682c-466a-4e8a-a9a4-23e936582ccd" />

### Expected Result:

- Estimated time should include a unit label (e.g., **“10 seconds”**)

<img width="1365" height="681" alt="Image" src="https://github.com/user-attachments/assets/710cbc2a-2a53-456c-b547-19ddbb6acb74" />

### Special Notes:

```
**Possible Root Cause:**

* UI formatting layer is missing time unit suffix in Confidential Exchange view
* Shared component between Public and Confidential flows is not applying localization/formatting rules consistently
* Backend may return only numeric value without metadata (unit type), and frontend fails to map it
```

### Workarounds:
None

### Repro Rate:
Always (10/10)

### Setup - OS - Browser - Environment

* OS / Devices: Windows 11 (x64), Motorola G23 (Android 14), Iphone XR (iOs 18.7.6)
* Screen resolution(s): 1366 x 768, 1600 x 720, 1792×828
* Browser(s) + version(s): Chrome (148.0.7778.97), Mozilla Firefox (147.0.2), Brave (1.86.148), Safari (18.7.4), Edge (145.0.3800.65)
* Wallet: Meteor
* App environment: Staging (trezu.app)
* Near Network: Mainnet

### Impact:

**User Impact:**

* Users cannot clearly interpret estimated time values
* Increases cognitive load and risk of misunderstanding timing expectations

**System Impact:**

[...]

Issue #625, opened 2026-05-07
[BUG] Flicker of "Confidential" state on page load/refresh


ori-near:
### Summary

When a logged-in user refreshes the dashboard, the UI briefly displays the public "Confidential" skeleton/shield view for a split second before the authenticated data populates.

**The Problem:**
- Even when authenticated, the app defaults to the "unauthorized" view during the initial load/handshake.
- This creates a jarring visual "pop" and makes the app feel slower or unauthenticated.

https://github.com/user-attachments/assets/1ec1d8a2-259d-4279-8fce-509d3ca00c59

**Expected Behavior:**
The app should maintain the authenticated state or show a generic loading spinner/transition that doesn't imply the user is "logged out" or "unauthorized" during the fetch.

### Steps To Reproduce

1. Log in to a Confidential Treasury.
2. Hard refresh the browser page.
3. Observe the UI showing shield icons/placeholders before the actual balances appear.

### Setup - OS - Browser - Environment

Mac OS - Chrome Version 147.0.7727.138 (Official Build)

### Priority

🟢 P3 : Normal

rknearn-qa:
Retested
Fixed

https://github.com/user-attachments/assets/2be5ddaf-4f6e-48f8-8025-4b600a4ba520

* OS / Devices: Windows 11 (x64), Motorola G23 (Android 14), Iphone XR (iOs 18.7.6)
* Screen resolution(s): 1366 x 768, 1600 x 720, 1792×828
* Browser(s) + version(s): Chrome (148.0.7778.97), Mozilla Firefox (147.0.2), Brave (1.86.148), Safari (18.7.4), Edge (145.0.3800.65)
* Wallet: Meteor
* App environment: Staging ([trezu.app](https://testenv.trezu.app/confidential-staging.sputnik-dao.near))
* Near Network: Mainnet

PR #643, opened 2026-05-13
fix: improve treasury load flow and exchange time formatting

Megha-Dev-19:
- Fixed a loading flicker where treasury pages could briefly show guest/confidential state before auth and treasury data finished loading. #625
- Enabled export button for confidential treasury dashboard. #629
- Fixed exchange request details to show estimated time with units (for example, `10 seconds`) instead of a raw number. #628


nt-fe/features/proposals/components/expanded-view/swap-expanded.tsx
```diff
@@ -93,9 +93,16 @@ function IntentsSwapExpanded({ data }: SwapExpandedProps) {
...
         expandableItems.push({
             label: t("estimatedTime"),
-            value: <span>{data.timeEstimate}</span>,
+            value: (
+                <span>
+                    {Number.isFinite(estimatedSeconds)
+                        ? t("seconds", { count: estimatedSeconds })
+                        : data.timeEstimate}
+                </span>
+            ),
```
frol: Have you considered using native browsers API `Intl.RelativeTimeFormat` for the relative time formatting?

frol (changes requested):

nt-fe/features/proposals/components/expanded-view/swap-expanded.tsx
Megha-Dev-19: but it's not relative, this time is considered after the user approves/executes the request.
basically how much time it will take for the deposit to be processed and then withdrawal/swap to initiate.

frol: @Megha-Dev-19 So there is `Intl.DurationFormat` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DurationFormat

frol: I don't understand the logic here now, what can the `timeEstimate` be, so it is warranted to display it as is it is not parsed as a number?

Megha-Dev-19: it's this time estimate, but we attach the 'seconds' string together for exchange request in proposal description`${proposalData.quote.timeEstimate} seconds`.
 
https://docs.near-intents.org/api-reference/oneclick/request-a-swap-quote#response-quote-time-estimate

frol (approved):

Issue #599, opened 2026-05-06
[BUG] Trezu - Dashboard: Funds are not reflected in Confidential Treasury balance after deposit


rknearn-qa:
### Summary

In the **Confidential Treasury Dashboard**, after receiving funds from a **Public Treasury**, the transaction appears in **Recent Transactions**, but the **total balance is not updated**. This creates inconsistency between transaction history and actual treasury balance.

### Preconditions:

- Confidential Treasury exists (https://testenv.trezu.app/confidential-staging.sputnik-dao.near/requests)
- Public Treasury exists (https://testenv.trezu.app/staging-qa.sputnik-dao.near)
- User can perform transfers between treasuries

### Steps To Reproduce

1. Send **1 NEAR / USDC / USDT** from Public Treasury to Confidential Treasury (https://testenv.trezu.app/staging-qa.sputnik-dao.near/requests/39)
2. Wait for transaction to be completed
3. Open **Confidential Treasury Dashboard**
4. Observe **Recent Transactions** and **total balance**

### Actual Result:

- Deposit is visible in **Recent Transactions**
- **Total balance is NOT updated**
- Funds are effectively not usable

<img width="1365" height="673" alt="Image" src="https://github.com/user-attachments/assets/c957fe37-c6d4-4ea4-9dc7-cea1960f0aed" />

<img width="1366" height="768" alt="Image" src="https://github.com/user-attachments/assets/926ec045-193f-41e4-a0c9-90c7f3c18921" />

https://github.com/user-attachments/assets/7e0b04f3-8eb4-45ee-b358-519d770c1ebb

### Expected Result:

- Deposit should be reflected in **Recent Transactions**
- **Total balance should increase accordingly**
- No discrepancy between transaction history and balance

### Special Notes:

```
Possible Root Cause

* Deposit indexing occurs, but **balance update logic is not triggered**
* Incorrect handling of **Public → Confidential transfer flow**
* Missing linkage between **transaction processing and balance aggregation**
* Possible requirement of deposit address vs account address not enforced

```

### Workarounds:
None

### Repro Rate:
Always (10/10)

### Setup - OS - Browser - Environment

* OS / Devices: Windows 11 (x64), Motorola G23 (Android 14), Iphone XR (iOs 18.7.6)
* Screen resolution(s): 1366 x 768, 1600 x 720, 1792×828
[...]

PR #647, opened 2026-05-13
fix: Improved token metadata resolution and simplify bridge asset matcher to address duplicate and missing networks on the list

Megha-Dev-19:
#650 #599

- Reworked backend token metadata flow to use clearer candidate normalization and source lookup order to avoid duplicate network display.
    1. `counterparties` table
    2. `near` / `wrap.near` fallback
    3. `tokens.json` (static canonical)
    4. Chaindefuser tokens API
    5. NearBlocks FT fallback (now normalizes `intents.near:` / `nep141:` before querying)
 - Renamed wrapper helper to clearer intent:
    - `fetch_tokens_with_defuse_extension` -> `fetch_tokens_metadata_enriched`

- Frontend:
    - Payment request: Missing receiver network fixed by matching networks using ids. (eg: in case of AURORA token, NEAR network was not visible).
    - Fixed ETH implicit address normalization in quote flow: mixed-case addresses were causing quote failures, so addresses are now normalized to lowercase before lookup.


frol (approved):
Overall, all these ad-hoc string manipulations do not bring peace of mind. It would be great to define the strict type for the asset type (and have an internal table with all the assets we support, and use their UUID as foreign key where applicable, including in the prices table). cc @ktarun1419 

That is the ideal state of things, but we are not there yet, so I am not going to block this PR since it addresses a real problem and unblocks us.

Issue #645, opened 2026-05-13
Improve the UX for informing users about what is confidential on the dashboard and which address the user is sending assets to.


olskik12:
**Dashboard**
On the dashboard page, add a “shield” icon to the Total Balance, Assets, and Recent Transactions sections with tooltips that explain to users that this data is confidential.

[Figma](https://www.figma.com/design/UlUkb3wL7fQq0xNvmjFnEg/TREZU?node-id=7554-44118&t=Syd5zI13ucQLfMbu-4)

<img width="1451" height="911" alt="Image" src="https://github.com/user-attachments/assets/f06ee84b-f809-4903-8339-2bb82251bee1" />

**Requests**
Add an icon and a tooltip before the address if the address is confidential.
<img width="1456" height="957" alt="Image" src="https://github.com/user-attachments/assets/ee8522cc-0bd8-49aa-81e7-c180da793733" />

On payment and exchange requests, do not show the name of the network in the details where the asset is stored directly in the UI. This information should be shown in a tooltip instead.
This should also be applied to the public treasury.

<img width="1459" height="960" alt="Image" src="https://github.com/user-attachments/assets/8437e492-d2ca-4a0f-bbb5-273a5bf00be7" />

PR #649, opened 2026-05-13
feat: Improved the UX for informing users about what is confidential on the dashboard and which address the user is sending assets to

Megha-Dev-19:
#645 
- Added shield icon and tooltip on dashboard sections (Total Balance, Assets, Recent Transactions) for confidential treasuries.
- Added shield icon before receiver in request list when destination is `near.com` in confidential treasury.
- Updated request/exchange amount displays to use hover tooltip for network info instead of inline network text.


nt-fe/features/proposals/components/expanded-view/transfer-expanded.tsx
```diff
@@ -109,11 +102,11 @@ export function TransferExpanded({ data }: TransferExpandedProps) {
...
             value: (
-                <Tooltip content={amountNetworkLabel}>
-                    <div>
-                        <Amount amount={data.amount} tokenId={data.tokenId} />
-                    </div>
-                </Tooltip>
+                <Amount
+                    amount={data.amount}
+                    tokenId={data.tokenId}
+                    showNetworkTooltip
```
frol, line 108: Should it be just `true` by default? Are there any places where we don't want to show the tooltip?

Megha-Dev-19: Yeah there are a few places where we use it like in form of payment request, exchange, there we want it in description.

frol (approved):

Issue #653, opened 2026-05-14
[BUG] Trezu - Payments: Destination Network incorrectly displayed as NEAR for Avalanche (Solana etc) payment request


rknearn-qa:
### Summary

When creating a payment request using **USDC on Avalanche network**, the resulting request details incorrectly display the **Destination Network** as **NEAR** instead of **Avalanche**. Although the user explicitly selects Avalanche as the recipient network during request creation, the final request metadata defaults to NEAR network in the UI.

### Preconditions:

- Payment request creation flow is available (https://testenv.trezu.app/confidential-staging.sputnik-dao.near/requests/59)
- User can create cross-network payment requests

### Steps To Reproduce

1. Start Payment request creation
2. Select **1 USDC (Avalanche Network)**
3. Enter recipient address:

   * `0xc366efeff224ea755118b19c30202869c5a85490`
4. Choose **Avalanche** as Recipient Network
5. Complete request creation
6. Open created request details
7. Observe **Destination Network** field

### Actual Result:

- Destination Network is displayed as: NEAR
- Selected Avalanche network is not reflected in request details

<img width="1365" height="679" alt="Image" src="https://github.com/user-attachments/assets/7e7647b1-7659-4cb1-a3ff-60fcec2b9e36" />
<img width="1085" height="478" alt="Image" src="https://github.com/user-attachments/assets/b662067d-c27d-4fb6-9717-e4f0630069d5" />

### Expected Result:

- Destination Network should display: Avalanche
- Request metadata should accurately reflect selected recipient network

### Special Notes:

```
**Possible Root Cause:**

* Destination network defaults to NEAR during request serialization or rendering
* Incorrect mapping between selected token network and displayed destination network
* Backend request payload may omit or override recipient network value
* UI component may fallback to treasury default network instead of selected destination network
```

### Workarounds:
None

### Repro Rate:
Always (10/10)

### Setup - OS - Browser - Environment

* OS / Devices: Windows 11 (x64), Motorola G23 (Android 14), Iphone XR (iOs 18.7.6)
* Screen resolution(s): 1366 x 768, 1600 x 720, 1792×828
* Browser(s) + version(s): Chrome (148.0.7778.97), Mozilla Firefox (147.0.2), Brave (1.86.148), Safari (18.7.4), Edge (145.0.3800.65)
* Wallet: Meteor
* App environment: Staging (https://testenv.trezu.app/confidential-staging.sputnik-dao.near/requests)
[...]

PR #655, opened 2026-05-14
fix: resolve confidential payment destination network from recipient and destination asset

Megha-Dev-19:
#653 
- Confirmed with Artur: near.com route is now applied only when `recipientType` is `CONFIDENTIAL_INTENTS`.
- In all other cases, destination network is derived from `destinationAsset`.
- Added back staging pill

### DIFF
--- a/nt-fe/components/page-component-layout.tsx
+++ b/nt-fe/components/page-component-layout.tsx
@@ -95,6 +95,15 @@ export function PageComponentLayout({
                 </div>
 
                 <div className="flex items-center gap-3">
+                    {isStaging && (
+                        <Pill
+                            title="Staging"
+                            icon={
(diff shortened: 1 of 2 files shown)


Megha-Dev-19:
@frol Is this check sufficient to derive the destination network from the quote? From my understanding, the only special case is near.com: we should use near.com only when recipientType is CONFIDENTIAL_INTENTS and the recipient is a valid NEAR address; otherwise we should use destinationAsset from the quote to resolve the network.

frol:
@Megha-Dev-19 Oh, that is hard question for me to answer from the top of my head since I have never really dived that deep into the quotes 🙁

Could you ask Artur if he remembers the details and could help us with his memory?

frol (approved):

PR #707, opened 2026-05-21
fix: migrate icon URLs to `near.com`

Megha-Dev-19:
#709


yurko-ai (approved):

akorchyn (approved):
The only question I have. Do we still have cases where dark differs from light ? If not maybe we might simplify it

Issue #671, opened 2026-05-14
[DEV] Implement Audit-Ready PDF Receipt / Payment Confirmation


olskik12:
Add the ability for users to generate PDF receipts / payment confirmations for all treasury types.

#634

ori-near:
Hi @frol – whenever you're back, please take a look at this ticket! Thank you

rknearn-qa:
Implemented

<img width="1365" height="682" alt="Image" src="https://github.com/user-attachments/assets/dc9aadd7-53e2-4f7e-8424-43772b3d07fe" />

* OS / Devices: Windows 11 (x64), Motorola G23 (Android 14), Iphone XR (iOs 18.7.6)
* Screen resolution(s): 1366 x 768, 1600 x 720, 1792×828
* Browser(s) + version(s): Chrome (149.0.7827.54), Mozilla Firefox (151.0.3), Brave (1.91.168), Safari (18.7.4), Edge (149.0.4022.52)
* Wallet: Meteor
* App environment: Staging ([trezu.app](https://testenv.trezu.app/confidential-staging.sputnik-dao.near/requests/83/receipt))
* Near Network: Mainnet

PR #713, opened 2026-05-21
feat: pdf receipt for payment, exchange

Megha-Dev-19:
#671 
- Added a new receipt page at `/{treasuryId}/requests/{id}/receipt` so users can open a transaction receipt, then print it or save it as PDF.
image

image

- Added two new database counters:
  - `receipts_generated` = how many receipts were created
  - `receipt_prints` = how many times users printed/saved receipts

- Fixed how the “executed on” date is picked in proposals table/view:
  - First use swap status time (`updatedAt`)
  - If that is missing, use the transaction timestamp
- Fixed missing USD amounts in executed payment/swap views:
  - If USD value is not present in proposal data, fetch it from the deposit-address quote API and show that instead


nt-fe/app/(treasury)/[treasuryId]/requests/[id]/receipt/page.tsx
```diff
@@ -0,0 +1,1055 @@
...
+                        <ReceiptValueSkeleton width="w-32" />
+                    </span>
+                ) : receiptDate.value ? (
+                    formatUserDate(receiptDate.value, {
+                        timezone: "UTC",
+                        includeTime: false,
+                    })
+                ) : (
+                    tCommon("notAvailable")
+                )}
```
Comment (frol) [inline nt-fe/app/(treasury)/[treasuryId]/requests/[id]/receipt/page.tsx]: The text should be localized and parameterized

nt-fe/app/(treasury)/[treasuryId]/requests/[id]/receipt/page.tsx
```diff
@@ -0,0 +1,1055 @@
...
+        <div className="space-y-6">
+            <div className="flex items-start justify-between border-b pb-4">
+                <div>
+                    <p className="text-xl font-medium">{title}</p>
+                    <p className="mt-1 text-sm text-muted-foreground">
+                        {tReceipt("generatedOn")}{" "}
+                        {formatUserDate(new Date(), {
+                            timezone: "UTC",
+                            includeTime: false,
+                        })}
```
Comment (frol) [inline nt-fe/app/(treasury)/[treasuryId]/requests/[id]/receipt/page.tsx]: The `generatedOn` must be parametrized with the date instead of concatenated outside of the localization as in other languages it may require to put the words after the date as well.

nt-be/migrations/20260521000002_add_receipt_metrics.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE usage_tracking
+    ADD COLUMN IF NOT EXISTS receipts_generated int4 NOT NULL DEFAULT 0,
+    ADD COLUMN IF NOT EXISTS receipt_prints int4 NOT NULL DEFAULT 0;
```
frol: Can we / should we align the naming? `receipt_generated` & `receipts_printed`?

frol (approved):
There are no blockers. LGTM

Consider addressing the review comments as part of this PR.

Issue #641, opened 2026-05-12
[Task] Updating e2e recent activity flow


ktarun1419:
### Background

The core values of the new system should be:

1. simplicity (let's avoid "clever" solutions that we would have a hard time operating and troubleshooting)
2. reliability (3-layer architecture )
3. recoverability (we should have a clear way to partially resync/reprocess the data in case of bugs)

### User Story

NA

### Acceptance Criteria

NA

frol:
After this issue is resolved, we should retest #595

ori-near:
@ktarun1419 reminder to please add a list of all the affected tickets that will be solved by this ticket

ktarun1419:
- [x]  #661 - This will show the accurate USD value.
- [ ]  #595 - This will fix the charts.
- [x]  - Fix: Correct executed date and creation date on the request page
- [x]  - Fix: Single row for exchange (Current : Shows 2 rows)
- [x]  - Fix: Missing recent history for the confidential
- [ ]  - Feat: Add refresh button and last updated in recent activity .
- [x]  - Feat: Link to proposal in recent history which created the transactions.
- [x]  - Feat: Add USD amount in the Recent transaction detail modal.

ktarun1419:
a quick update here 
This is blocked from the defuse team side, and they said they cannot confirm the date when the issue will be fixed 

We are planning to work on a workaround, until then I am moving this to blocked

rknearn-qa:
Implemented for Confidential Treasuries as well

Move it to Done

PR #721, opened 2026-05-24
refactor: Upgrade confidential history indexing pipeline to be reliable and scalable

ktarun1419:
Creates a new pipeline for the confidential history.

#641


nt-be/src/handlers/balance_changes/confidential_list.rs
```diff
@@ -0,0 +1,746 @@
...
+
+use bigdecimal::{BigDecimal, FromPrimitive, Zero};
+use chrono::{DateTime, Utc};
+use sqlx::{PgPool, QueryBuilder};
+
+use crate::AppState;
+use crate::handlers::token::{TokenMetadata, fetch_tokens_with_fallback};
+use crate::routes::{BalanceChangesQuery, EnrichedBalanceChange, SwapInfo};
+
+const SENT_LEG_TO_ACCOUNT_EXPR: &str = "regexp_replace(recipient, '^near:', '')";
```
frol: Instead of stripping / replacing the "near:" prefix, should we embrace it and always use "<chain_id>:<account_id>" format everywhere? And so define the strict Rust type "NetworkAccountId" instead of "String" that would enforce it?

nt-be/src/handlers/balance_changes/confidential_list.rs
```diff
@@ -0,0 +1,746 @@
...
+
+#[derive(Debug, sqlx::FromRow)]
+struct ConfidentialBalanceChangeRow {
+    id: i64,
+    history_event_id: i64,
+    dao_id: String,
```
frol: Since #617, we should use AccountId type for the `dao_id`

nt-be/src/handlers/balance_changes/confidential_list.rs
```diff
@@ -0,0 +1,746 @@
...
+    transaction_type: String,
```
frol: Should/can it be enum?

nt-be/src/handlers/balance_changes/confidential_list.rs
```diff
@@ -0,0 +1,746 @@
...
+    amount_in: Option<BigDecimal>,
+    amount_out: BigDecimal,
+    amount_in_usd: Option<BigDecimal>,
+    amount_out_usd: Option<BigDecimal>,
+    origin_balance_before: Option<BigDecimal>,
+    origin_balance_after: Option<BigDecimal>,
+    destination_balance_before: Option<BigDecimal>,
+    destination_balance_after: Option<BigDecimal>,
+    recipient: String,
+    counterparty: String,
```
frol, line 47: In which case the `recipient` and `counterparty` are different?

nt-be/src/handlers/balance_changes/confidential_list.rs
```diff
@@ -0,0 +1,746 @@
...
+        for change in &mut enriched {
+            change.token_metadata = metadata_map.get(&change.token_id).cloned();
+        }
+    }
+
+    for (change, leg) in enriched.iter_mut().zip(leg_rows.iter()) {
+        if let Some(swap) = leg.build_swap_info(&metadata_map) {
+            change.swap = Some(swap);
+        }
+    }
```
frol: Why `to_enriched` implementation does not do it internally?

nt-be/src/handlers/balance_changes/confidential_list.rs
```diff
@@ -0,0 +1,746 @@
...
+
+    if params.include_prices.unwrap_or(false) {
+        attach_prices(state, &mut enriched).await;
+    }
```
frol: I feel that in the context of executed transactions, the price must equal to `amount_out_usd / (amount_in_usd - fee_in_usd)`, otherwise our USD estimates are disconnected from the price and that would lead to odd results.

Do you think otherwise?

nt-be/src/handlers/balance_changes/confidential_list.rs
```diff
@@ -0,0 +1,746 @@
...
+fn resolve_swap_metadata(
+    token_id: &str,
+    metadata_map: &HashMap<String, TokenMetadata>,
+) -> TokenMetadata {
+    if let Some(meta) = metadata_map.get(token_id) {
+        return meta.clone();
+    }
+    let symbol = token_id
+        .split('.')
+        .next()
```
frol: We need to stop doing such manipulations on the fly. Let's have the gold data source for the tokens metadata.

nt-be/src/handlers/balance_changes/goldsky_enrichment.rs
```diff
@@ -527,6 +534,185 @@ async fn get_monitored_accounts(
...
+            .get("SuccessValue")
+            .and_then(|v| v.as_str())
+            .map(ToString::to_string)
+    } else {
+        let marker_idx = status.find("SuccessValue")?;
+        let rest = &status[marker_idx..];
+        let first_quote = rest.find('"')?;
+        let after_first_quote = &rest[first_quote + 1..];
+        let second_quote = after_first_quote.find('"')?;
+        Some(after_first_quote[..second_quote].to_string())
```
frol: What is going on here? Why is it needed?

nt-be/src/handlers/intents/confidential/balance_changes_projector/classify.rs
```diff
@@ -0,0 +1,531 @@
...
+        &dao_id,
+        &recipient,
+        origin_asset.as_deref(),
+        &destination_asset,
+    ) else {
+        return Ok(None);
+    };
+
+    let amount_out = parse_decimal(
+        payload_str(&row.raw_payload, "amountOutFormatted"),
```
frol: I feel that you underutilize serde derive macros and overuse serde_json::Value type. It is really hard to review all these parsing calls.

https://serde.rs/

nt-be/src/handlers/intents/confidential/history_store/linking.rs
```diff
@@ -0,0 +1,116 @@
...
+        candidate AS (
+            SELECT he.id AS history_event_id
+            FROM confidential_history_events he
+            JOIN intent i
+              ON he.account_id = i.dao_id
+             AND he.deposit_address = i.deposit_address
+             AND (
+                 i.recipient IS NULL
+                 OR he.recipient = i.recipient
+                 OR CONCAT('near:', he.recipient) = i.recipient
```
frol: I really-really don't like these "near:" hardcodings randomly scattered all over the code

nt-be/src/handlers/proposals/get_proposals.rs
```diff
@@ -329,8 +343,29 @@ async fn enrich_confidential_proposals(proposals: &mut [Proposal], pool: &PgPool
...
+            ci.status,
+            ci.correlation_id,
+            ci.notes,
+            ci.proposal_created_at,
+            ci.executed_at,
+            gold.amount_in_usd::TEXT AS gold_amount_in_usd,
+            gold.amount_out_usd::TEXT AS gold_amount_out_usd,
+            gold.usd_change::TEXT AS gold_usd_change
+        FROM confidential_intents ci
+        LEFT JOIN LATERAL (
```
frol, line 360: It looks like we have a mix of gold and broze data sources here.

I suggest we avoid mixing those.

nt-be/src/handlers/treasury/creation_status.rs
```diff
@@ -5,7 +5,7 @@ use near_api::{NearToken, Tokens};
 use reqwest::StatusCode;
 use serde::Serialize;
 
-pub const LOW_BALANCE_THRESHOLD: NearToken = NearToken::from_near(5);
+pub const LOW_BALANCE_THRESHOLD: NearToken = NearToken::from_near(0);
```
frol: Why?

nt-be/src/handlers/intents/confidential/history_store/linking.rs
```diff
@@ -0,0 +1,116 @@
...
+              AND status = 'submitted'
+              AND NOT EXISTS (
+                  SELECT 1
+                  FROM confidential_intents existing
+                  WHERE existing.history_event_id = $1
+              )
+              AND (
+                  $4::TEXT IS NULL
+                  OR quote_metadata->'quoteRequest'->>'recipient' = $4
+                  OR quote_metadata->'quoteRequest'->>'recipient' = CONCAT('near:', $4)
```
frol: ^ (marking all "near:" places, so we don't forget about them)

nt-be/src/handlers/intents/confidential/balance_snapshots/worker.rs
```diff
@@ -0,0 +1,290 @@
...
+    SnapshotRow, insert_snapshot_rows, latest_snapshot_at, load_latest_balances_per_asset,
+};
+use crate::AppState;
+use crate::constants::intents_tokens::get_defuse_tokens_map;
+use crate::handlers::intents::confidential::balances::fetch_confidential_balances;
+use crate::handlers::intents::confidential::history_store::load_confidential_history_accounts;
+
+pub const HOURLY_SNAPSHOT_CRON_TICK_SECS: u64 = 3600;
+
+const SNAPSHOT_DEDUP_WINDOW_SECS: i64 = 3300;
```
frol: Since `std::time::Duration::from_secs` and other constructor methods are `const fn`, let's use the `Duration` type instead of `_SECS: i64` (apply everywhere as there are dozen of places)

frol (changes requested):
This is a massive PR! Thank you for authoring this major revamp.

I have hard time understanding the structure of all the workers and their dependencies. Let's use `bronze_`, `silver_`, and `gold_` prefixes or folder names to clearly communicate the layers.

Let's make sure that the gold level can be used without complex queries of CASE, CONCAT, JOINs, etc.

Let's make use of strict types and avoid pure "&str", String, and integers.

nt-be/src/handlers/balance_changes/confidential_list.rs
ktarun1419: 1. In the deposit, the `recipient` is the `dao_id`, and the `counterparty` is the `intents.near`
2. In exchange as well, the `recipient` is the `dao_id`, and the `counterparty` is the `intents.near`

nt-be/src/handlers/balance_changes/confidential_list.rs
ktarun1419: `to_enriched` runs before `metadata_map` is fetched, and `build_swap_info` needs that map, so it's currently a post-pass. We can fold it in by computing  `metadata_map` first and passing it into `to_enriched`.

but will make it better by computing `metadata_map` before enriching (it only depends on `leg_rows`, not on enriched), then let `to_enriched` take the map and do everything in one pass, collapsing all three loops

nt-be/src/handlers/intents/confidential/balance_snapshots/worker.rs
ktarun1419: fixed

nt-be/src/handlers/intents/confidential/history_store/linking.rs
ktarun1419: using bare account everywhere 
removed these on-the-go manipulations

nt-be/src/handlers/intents/confidential/history_store/linking.rs
ktarun1419: fixed

nt-be/src/handlers/balance_changes/confidential_list.rs
ktarun1419: fixed

nt-be/src/handlers/balance_changes/confidential_list.rs
ktarun1419: yes, implemented it

nt-be/src/handlers/balance_changes/confidential_list.rs
ktarun1419: fixed

nt-be/src/handlers/balance_changes/confidential_list.rs
ktarun1419: For executed swaps, use the executed quote USD values consistently. Don’t mix amount_out_usd with external spot prices.
But amount_out_usd / (amount_in_usd - fee_in_usd) is a net execution ratio, not a per-token price.

nt-be/src/handlers/balance_changes/goldsky_enrichment.rs
ktarun1419: to parse the proposal_id

nt-be/src/handlers/proposals/get_proposals.rs
ktarun1419: for our new system 
- both confidential_intents and gold_confidential_history_events comes under gold category

nt-be/migrations/20260603000001_confidential_history_pipeline.sql
```diff
@@ -0,0 +1,138 @@
...
+    raw_payload JSONB NOT NULL,
+    updated_at TIMESTAMPTZ NOT NULL DEFAULT NOW()
+);
+
+CREATE UNIQUE INDEX IF NOT EXISTS idx_bche_unique_event
+    ON bronze_confidential_history_events (account_id, created_at_external, deposit_address);
+CREATE INDEX IF NOT EXISTS idx_bche_in_flight
+    ON bronze_confidential_history_events (account_id, status) WHERE status IN ('SUCCESS');
+CREATE INDEX IF NOT EXISTS idx_bche_gold_scan
+    ON bronze_confidential_history_events (account_id, status, created_at_external, id);
```
frol, line 24: This index looks odd to me since it contains `id` column. Are you sure it is a useful index? (I see other indexes also include `id` column, so review them as well)

nt-be/migrations/20260603000001_confidential_history_pipeline.sql
```diff
@@ -0,0 +1,138 @@
...
+
+
+CREATE TABLE IF NOT EXISTS gold_confidential_history_events (
+    id                          BIGSERIAL PRIMARY KEY,
+    history_event_id            BIGINT NOT NULL UNIQUE
+        REFERENCES bronze_confidential_history_events(id) ON DELETE CASCADE,
+    intent_id                   INTEGER REFERENCES confidential_intents(id),
+    dao_id                      TEXT NOT NULL,
+    transaction_type            TEXT NOT NULL
+        CHECK (transaction_type IN ('sent', 'exchange', 'deposit')),
```
frol: Have you considered using enums instead of strings? The benefit is that it will save on storage: https://stackoverflow.com/questions/76465657/how-do-i-create-custom-postgres-enum-types-in-rust-sqlx

nt-be/migrations/20260603000001_confidential_history_pipeline.sql
```diff
@@ -0,0 +1,138 @@
...
+    origin_asset                TEXT,
+    destination_asset           TEXT NOT NULL,
```
frol, line 62: I had previously mentioned that it could be a good idea to extract assets into a separate table and reference them by an internal integer id instead of strings. Do you think it could be a good idea?

nt-be/migrations/20260603000001_confidential_history_pipeline.sql
```diff
@@ -0,0 +1,138 @@
...
+    destination_balance_before  NUMERIC,
+    destination_balance_after   NUMERIC,
+    recipient                   TEXT NOT NULL,
+    refund_to                   TEXT NOT NULL,
+    counterparty                TEXT NOT NULL,
+    deposit_address             TEXT NOT NULL,
+    deposit_memo                TEXT,
+    block_height                BIGINT,
+    block_time                  TIMESTAMPTZ,
+    transaction_hash            TEXT,
```
frol: Is my understanding correct that these will be NULL for deposits and non-NULL for payments and only reference the confidential intent signing approval transaction, and not the payment itself? Or are these confidential network block/transaction? (if so, I wonder if there is any value in having them in our gold projection)

The naming of these columns is too generic. It might be worth prefixing them with `proposal_approved_at_...`

nt-be/migrations/20260603000001_confidential_history_pipeline.sql
```diff
@@ -0,0 +1,138 @@
...
+    quote_created_at            TIMESTAMPTZ NOT NULL,
+    proposal_created_at         TIMESTAMPTZ,
+    executed_at                 TIMESTAMPTZ,
+    created_at                  TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+    updated_at                  TIMESTAMPTZ NOT NULL DEFAULT NOW()
```
frol, line 83: 1. When `updated_at` will be different from `created_at`?
2. Currently, internal indexing metadata (`created_at` and `updated_at`) is mixed with the business logic metadata (`executed_at`) - I think it is worth coming up with more descriptive name for `executed_at` or come up with a convention for the internal names like `created_at` (in Python world, devs use one or two underscores prefix).

nt-be/migrations/20260603000001_confidential_history_pipeline.sql
```diff
@@ -0,0 +1,138 @@
...
+CREATE INDEX IF NOT EXISTS idx_gchpe_errors_dao
+    ON gold_confidential_history_projection_errors (dao_id, updated_at DESC);
+
+
+CREATE TABLE IF NOT EXISTS gold_confidential_balance_snapshots (
+    dao_id        TEXT        NOT NULL,
+    asset         TEXT        NOT NULL,
+    snapshot_at   TIMESTAMPTZ NOT NULL,
+    raw_balance   NUMERIC     NOT NULL,
+    balance       NUMERIC     NOT NULL,
```
frol, line 129: I'd leave it up to the user to normalize the balance based on the asset metadata decimals. What is the benefit of having `raw_balance` + `balance` on the database level?

nt-be/src/handlers/balance_changes/goldsky_enrichment.rs
```diff
@@ -527,6 +534,192 @@ async fn get_monitored_accounts(
...
+    .bind(proposal_created_at)
+    .execute(app_pool)
+    .await?;
+
+    if result.rows_affected() > 0 {
+        if let Some(history_event_id) =
+            link_intent_to_history_event(app_pool, dao_id, payload_hash).await?
+        {
+            log::info!(
+                "[goldsky-enrichment] linked proposal intent {}/{} to history_event_id={}",
```
frol: We should consider using `tracing` instead of `log` crate. Tracing has a much bigger ecosystem of plugins and has `target` attribute (in most cases it auto-detects the context from the source code), so you won't need to manually add `[goldsky-enruchment]`, and we should also consider adding Sentry integration (we already use it for frontend errors).

nt-be/src/handlers/intents/confidential/gold/history_events/repository.rs
```diff
@@ -0,0 +1,373 @@
...
+    .bind(dao_id)
+    .bind(reason)
+    .bind(raw_payload)
+    .execute(&mut **tx)
+    .await?;
+
+    Ok(())
+}
+
+pub(crate) async fn delete_stale_gold_rows(
```
frol, line 357: Please, add a comment explaining why this function is needed (deleting from the gold table should be a rare or ideally impossible event, I feel).

nt-be/src/handlers/intents/confidential/types/account_id.rs
```diff
@@ -0,0 +1,98 @@
+//! Bare account strings for NEAR Intents / 1Click confidential flows.
+//!
+//! Bronze, gold, and `quote_metadata` store addresses without `chain:` prefixes
+//! (`bob.near`, `0xabc…`, `intents.near`).
+
+use std::fmt;
+
+fn looks_like_near_account(value: &str) -> bool {
+    value.ends_with(".near") || value.ends_with(".testnet") || value.ends_with(".tg")
```
frol, line 9: There are also `0s`, `0x`, and 64-hex implicit account ids. Is this relevant here? You may also use `AccountId::parse` to verify if the account id is valid NEAR account or not.

nt-be/src/handlers/balance_changes/confidential_list.rs
frol: @ktarun1419 Well, the quote price is a promise, but the user needs the price of the actual swap I feel. At least, for the tax purposes that would be the value they need, otherwise there will be a discrepancy. Do you think there is a value in showing the quote price?

nt-be/src/handlers/balance_changes/goldsky_enrichment.rs
```diff
@@ -527,6 +534,192 @@ async fn get_monitored_accounts(
...
+                .filter(|s| !s.is_empty())
+                .map(ToString::to_string);
+        }
+
+        let inner = status
+            .strip_prefix("SuccessValue(")?
+            .strip_suffix(')')?
+            .trim();
+        let encoded = inner.strip_prefix('"')?.strip_suffix('"')?.trim();
+        (!encoded.is_empty()).then(|| encoded.to_string())
```
frol, line 555: In which case the `serde_json::from_str` would not be sufficient? I don't understand where this `SuccessValue("")` comes from

nt-be/src/handlers/balance_changes/goldsky_enrichment.rs
frol: Follow up: https://github.com/NEAR-DevHub/trezu/pull/721#pullrequestreview-4456269342

nt-be/src/handlers/balance_changes/goldsky_enrichment.rs
ktarun1419: `serde_json::from_str` is sufficient for the current normalized Goldsky rows, e.g. `{"SuccessValue":"MTM2"}` and `{"SuccessValue":""}`.

The fallback handles the raw NEAR execution status rendering that we had in the Goldsky/gap-filling data, e.g. `SuccessValue("MTE=")`. That shape is documented in `docs/sponsor-call-pairs-and-gap-filling.md` for `add_proposal` outcomes.

`SuccessValue("")` comes from successful calls that return no value, for example `act_proposal`. The useful case is `add_proposal`, where the success value is a base64-encoded proposal_id.

nt-be/migrations/20260603000001_confidential_history_pipeline.sql
ktarun1419: For `idx_bche_gold_scan`, `id` is intentional because the Gold projector scans Bronze rows with:

`WHERE account_id = ? AND status = 'SUCCESS' AND created_at_external >= ?`
`ORDER BY created_at_external ASC, id ASC`

So `id` is not used for filtering, but it gives deterministic replay order when multiple 1Click history rows have the same `created_at_external`

Some Gold indexes using `created_at` look less aligned with the current read path. I’ll clean those up in a follow-up change

nt-be/src/handlers/intents/confidential/types/account_id.rs
ktarun1419: The current suffix check is only used for our current confidential DAO signer ids are all named DAO accounts like `*.sputnik-dao.near`

nt-be/migrations/20260603000001_confidential_history_pipeline.sql
ktarun1419: `raw_balance` is the source value from 1Click, in base units. `balance` is a derived decimal value used by the read path.

The benefit of storing both is that we keep the exact upstream value for audit, while avoiding metadata lookups and repeated normalization

nt-be/migrations/20260603000001_confidential_history_pipeline.sql
ktarun1419: `updated_at` differs from `created_at` when the projection row is replayed or enriched after first insert. For ex: proposal/execution metadata can arrive later and update the existing Gold row.

Agree names are too generic, will update them

nt-be/migrations/20260603000001_confidential_history_pipeline.sql
ktarun1419: Yes, for deposits, there is no DAO/v1.signer approval transaction, so these fields will be NULL

They are useful for links/dates in the UI, but I agree they should be named more clearly.

nt-be/migrations/20260603000001_confidential_history_pipeline.sql
ktarun1419: Yes, I think that is a good idea.

With a dedicated assets table, we would have more control over asset handling. We could centralize metadata, consistently filter out unsupported assets, and add or remove supported assets in one place. we can refer to that table everywhere bronze,gold and frontend as well.


Should I implement it in this PR or pick this up later?

nt-be/src/handlers/intents/confidential/balance_changes_projector/classify.rs
ktarun1419: I acknowledged it and used it in as many places as possible

nt-be/src/handlers/balance_changes/confidential_list.rs
ktarun1419: No, we are not using the quote prices 
We are using the finalised prices and amounts returned from the history api

nt-be/src/handlers/balance_changes/goldsky_enrichment.rs
ktarun1419: I have implemented the tracing for this history pipeline
Will implement at the global level in the backend in another PR

nt-be/migrations/20260603000001_confidential_history_pipeline.sql
frol: > Should I implement it in this PR or pick this up later?

@ktarun1419 If it is easier to implement now than to migrate later, let's implement it now. Otherwise, let's come back to it later.

ktarun1419: @frol It will take time to implement it
Let's roll out the confidential pipeline as it's getting larger already. We can revisit this after/with the public pipeline and will add migrations (we will need them for the other tables as well)

nt-be/src/handlers/intents/confidential/gold/history_events/repository.rs
ktarun1419: added

frol (approved):

frol:
All my review comments are fully addressed. Let's ship it to staging!

Issue #730, opened 2026-05-26
[DEV] Add metric for Address Book transactions


Megha-Dev-19:
### Description:
Add a metric to track transactions where the recipient is selected from the Address Book.
This will help us measure Address Book usage and understand adoption over time.

### Scope:
- Track whether a transaction recipient came from Address Book selection.

rknearn-qa:
Implemented

<img width="1365" height="689" alt="Image" src="https://github.com/user-attachments/assets/710d465d-cfd3-4812-8560-60e067c6a61a" />

* OS / Devices: Windows 11 (x64), Motorola G23 (Android 14), Iphone XR (iOs 18.7.6)
* Screen resolution(s): 1366 x 768, 1600 x 720, 1792×828
* Browser(s) + version(s): Chrome (148.0.7778.97), Mozilla Firefox (147.0.2), Brave (1.86.148), Safari (18.7.4), Edge (145.0.3800.65)
* Wallet: Meteor
* App environment: Staging (trezu.app)
* Near Network: Mainnet

PR #731, opened 2026-05-26
feat: track payment proposals created from address book recipients

Megha-Dev-19:
#730

### DIFF
--- a/nt-be/src/handlers/relay/submit.rs
+++ b/nt-be/src/handlers/relay/submit.rs
@@ -556,39 +556,39 @@ pub async fn relay_delegate_action(
                     // gas_covered_transactions fires for every relayed action.
                     // The proposal metric only fires when proposalType is explicitly provided.
                     let proposal_metric = match request.proposal_type.as_deref() {
-                        Some("swap") => "swap_proposals",
-                        Some("payment") => "payment_proposals",
-                        Some("vote") => "votes_casted",
-                        Some(_) => "other_proposals_submitted",
-                        None => "",
+                        Some("swap") => {
+                            Some(crate::services::platform_metrics::PlatformMetric::SwapProposals)
+                        }
+                        Some("payment") => {
+                            Some(crate::services::platform_metrics::PlatformMetric::PaymentProposals)
+                        }
+                        Some("vote") => {
+                            Some(crate::services::platform_metrics::PlatformMetric::VotesCasted)
+                        }
+                        Some(_) => Some(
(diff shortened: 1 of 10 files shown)


nt-be/src/handlers/relay/submit.rs
```diff
@@ -558,7 +562,19 @@ pub async fn relay_delegate_action(
...
                     };
-                    if proposal_metric.is_empty() {
+                    let mut metrics = vec!["gas_covered_transactions"];
+                    if !proposal_metric.is_empty() {
+                        metrics.push(proposal_metric);
+                    }
+                    if request.address_book_payment
+                        && request.proposal_type.as_deref() == Some("payment")
+                    {
+                        metrics.push("address_book_payment_proposals");
```
ktarun1419: Small suggestion: can we use an enum for these metrics instead of raw strings? Since they map to DB columns, it would avoid typos and give us compile-time safety.

Example:

enum PlatformMetric {
    PaymentProposals,
    AddressBookPaymentProposals,
    ...
}

impl PlatformMetric {
    fn column(&self) -> &'static str {
        match self {
            Self::PaymentProposals => "payment_proposals",
            Self::AddressBookPaymentProposals => "address_book_payment_proposals",
             ....
        }
    }
}

it can push PlatformMetric::AddressBookPaymentProposals instead of "address_book_payment_proposals".

nt-fe/app/(treasury)/[treasuryId]/payments/components/payment-form-section.tsx
```diff
@@ -219,6 +221,10 @@ export function PaymentFormSection<
         }
     }, [selectedContact, setRecipientValue]);
 
+    useEffect(() => {
```
Comment (ktarun1419) [inline nt-fe/app/(treasury)/[treasuryId]/payments/components/payment-form-section.tsx:224]: this may count manually typed recipients as address book payments if the address matches an address book entry. Should this metric track only explicit address book selections, or any recipient that exists in the address book?

ktarun1419 (approved):

nt-fe/app/(treasury)/[treasuryId]/payments/components/payment-form-section.tsx
Comment (Megha-Dev-19) [reply nt-fe/app/(treasury)/[treasuryId]/payments/components/payment-form-section.tsx:224]: Good point. My reasoning was that users usually won’t manually type a saved address, since the Requestor role is the one managing the address book.

But even then, if a user selects an address book entry, edits it, and later changes it back to the same saved value, I still want to capture that as address-book usage.

Megha-Dev-19:
Note: I’ll look into the failing test in a separate PR. I’ve confirmed it is not related to the current changes.

Issue #704, opened 2026-05-20
[Task] Show Exchange Fee and Tooltip on the Full Request Page


ori-near:
### Background

Right now, when a user looks at the full details of an executed request, the exchange fee is missing:
<img width="1625" height="926" alt="Image" src="https://github.com/user-attachments/assets/431cdaac-b131-4407-9e39-74c16d5f5578" />

 We need to display the exact fee and the explanation tooltip just like we do in the pre-trade review screen, so users know exactly what they were charged and why.

<img width="905" height="874" alt="Image" src="https://github.com/user-attachments/assets/aba5962e-71de-46e4-8e8d-62b214779c4b" />

### User Story

As a user, I want to clearly see the exchange fee amount and an explanation tooltip on the executed request page, so I understand how much I paid and what the fee covers.

### Acceptance Criteria

- [ ] Add a row labeled "Exchange Fee" to the expanded details section of the executed request page.
- [ ] Display the fee value as both a percentage and the exact token amount (for example: 0.7% / 0.035 USDC).
- [ ] Include the small info icon (i) next to the "Exchange Fee" label.
- [ ] When a user hovers over the info icon, it must show the exact same tooltip used in the review step: "The 0.70% fee incurred here covers the NEAR Intents protocol costs to facilitate your trade and the Trezu widget fee. This amount is automatically calculated into your quoted rate."

rknearn-qa:
Implemented

<img width="1359" height="676" alt="Image" src="https://github.com/user-attachments/assets/08edd202-6867-4fce-988e-afa4b67a3431" />

* OS / Devices: Windows 11 (x64), Motorola G23 (Android 14), Iphone XR (iOs 18.7.6)
* Screen resolution(s): 1366 x 768, 1600 x 720, 1792×828
* Browser(s) + version(s): Chrome (148.0.7778.97), Mozilla Firefox (147.0.2), Brave (1.86.148), Safari (18.7.4), Edge (145.0.3800.65)
* Wallet: Meteor
* App environment: Staging ([trezu.app](https://testenv.trezu.app/staging-qa.sputnik-dao.near/requests))
* Near Network: Mainnet

PR #732, opened 2026-05-26
feat: added exchange fees on request page

ktarun1419:
#704


nt-fe/app/(treasury)/[treasuryId]/exchange/page.tsx
Comment (Megha-Dev-19) [inline nt-fe/app/(treasury)/[treasuryId]/exchange/page.tsx:672]: we already had a key for it in translations, can we reuse that

Megha-Dev-19 (approved):

nt-fe/app/(treasury)/[treasuryId]/exchange/page.tsx
Comment (ktarun1419) [reply nt-fe/app/(treasury)/[treasuryId]/exchange/page.tsx:672]: I kept this translation separate intentionally so all request-detail display strings stay in one namespace. It also gives us the flexibility to change the request-detail copy later without affecting the exchange review modal.

Comment (Megha-Dev-19) [reply nt-fe/app/(treasury)/[treasuryId]/exchange/page.tsx:672]: Let's keep same, as the description will remain same, I think we have a common key or something were we have such things, it not we can create one and keep it there.
It doesn't make sense to have same values in multiple places.

Comment (ktarun1419) [reply nt-fe/app/(treasury)/[treasuryId]/exchange/page.tsx:672]: make sense. 
Updated

Issue #724, opened 2026-05-25
[DEV] Change NEAR asset description and naming


olskik12:
In the dashboard, in the Assets section, rename “Wrapped NEAR fungible token” to “Near”

<img width="795" height="117" alt="Image" src="https://github.com/user-attachments/assets/7898a573-2290-4204-abba-9ff45f2235f2" />

Also, in the Deposit flow modal, we should rename “Near Protocol” to “NEAR,” since we don’t use that name anywhere else and it’s inconsistent.

<img width="558" height="377" alt="Image" src="https://github.com/user-attachments/assets/734aaf48-6a0c-495e-93d8-2aaefdbba2c7" />

olskik12:
Hi @Megha-Dev-19 @ktarun1419 Can you please take a look at this ?

rknearn-qa:
Duplicates #676

rknearn-qa:
Retested
Fixed

<img width="1236" height="687" alt="Image" src="https://github.com/user-attachments/assets/31544b4d-2697-45e3-aded-47c8c31ff7d5" />
<img width="1360" height="694" alt="Image" src="https://github.com/user-attachments/assets/e2cfd80a-448f-4386-a87b-aa6baabf0e2a" />

* OS / Devices: Windows 11 (x64), Motorola G23 (Android 14), Iphone XR (iOs 18.7.6)
* Screen resolution(s): 1366 x 768, 1600 x 720, 1792×828
* Browser(s) + version(s): Chrome (148.0.7778.97), Mozilla Firefox (147.0.2), Brave (1.86.148), Safari (18.7.4), Edge (145.0.3800.65)
* Wallet: Meteor
* App environment: Staging (trezu.app)
* Near Network: Mainnet

Issue #723, opened 2026-05-25
[DEV] Incorrect network name on the request details page for the “Destination Network” row


olskik12:
When the user makes a payment to a confidential wallet or treasury address, the network in the current UI differs from the one the user selected during creation.

<img width="1386" height="404" alt="Image" src="https://github.com/user-attachments/assets/a59d393b-3a14-4bb2-8475-3b2d976fbad4" />

We should simply display “near.com” in the row, because this causes confusion and inconsistency. The user selects one network, but we show a different one in the UI

olskik12:
@Megha-Dev-19 @ktarun1419 Could you please take a look at this issue when you have time?

rknearn-qa:
Retested
Fixed: near.com network is displayed

<img width="1365" height="670" alt="Image" src="https://github.com/user-attachments/assets/08a02296-4777-42a3-86a9-1c022b100a8a" />

* OS / Devices: Windows 11 (x64), Motorola G23 (Android 14), Iphone XR (iOs 18.7.6)
* Screen resolution(s): 1366 x 768, 1600 x 720, 1792×828
* Browser(s) + version(s): Chrome (148.0.7778.97), Mozilla Firefox (147.0.2), Brave (1.86.148), Safari (18.7.4), Edge (145.0.3800.65)
* Wallet: Ledger
* App environment: Staging ([trezu.app)](https://testenv.trezu.app/roma-kucher-qa-ledger.sputnik-dao.near/requests)
* Near Network: Mainnet

Issue #711, opened 2026-05-21
[BUG] Trezu - Network: Aptos network icon has inconsistent shape compared to other network icons


rknearn-qa:
### Summary

In the Payments/Exchange flows, the **Aptos** network icon is displayed with a square shape while other network icons use rounded styling. This creates a visual inconsistency in the network selection UI, especially noticeable in Dark Theme.

### Preconditions:

- Treasury is created
- Payments or Exchange flow is accessible
- Asset supporting Aptos network is available (e.g., USDT on Aptos)
- Dark Theme is enabled

### Steps To Reproduce

1. Open Payments or Exchange page
2. Select asset with Aptos network support (e.g., USDT)
3. Enable Dark Theme
4. Observe Aptos network icon appearance

### Actual Result:

- Aptos network icon appears with square shape/corners
- Other network icons are displayed with rounded styling
- UI appearance is visually inconsistent

<img width="1357" height="688" alt="Image" src="https://github.com/user-attachments/assets/0e2d1bc2-031f-4a50-9487-6d8af7b743b6" />

### Expected Result:

- Aptos network icon should follow the same rounded styling as other network icons
- Network icon presentation should remain visually consistent across all supported chains/themes

### Special Notes:

```
**Possible Root Cause:**

* Aptos icon asset lacks border-radius styling
* Icon component applies different mask/container styling for Aptos asset
* Inconsistent icon source formatting (SVG/PNG dimensions/background)
* Dark Theme styling exposes missing icon normalization
```

### Workarounds:
None

### Repro Rate:
Always (10/10)

### Setup - OS - Browser - Environment

* OS / Devices: Windows 11 (x64), Motorola G23 (Android 14), Iphone XR (iOs 18.7.6)
* Screen resolution(s): 1366 x 768, 1600 x 720, 1792×828
* Browser(s) + version(s): Chrome (148.0.7778.97), Mozilla Firefox (147.0.2), Brave (1.86.148), Safari (18.7.4), Edge (145.0.3800.65)
* Wallet: Meteor
* App environment: Staging ([trezu.app](https://testenv.trezu.app/staging-qa.sputnik-dao.near))
* Near Network: Mainnet

### Impact:

**User Impact:**

* Minor visual inconsistency in network selection UI
* Reduces polish and design consistency

**System Impact:**

* Network icon styling is not standardized across supported chains

**Business / Release Risk:**

* Low risk — cosmetic/UI consistency issue only
[...]

Issue #676, opened 2026-05-15
[BUG] Trezu - Dashboard: Native NEAR token is displayed as Wrapped NEAR fungible token


rknearn-qa:
### Summary

On the **Dashboard → Assets** page, the native **NEAR** token is incorrectly labeled as **“Wrapped NEAR fungible token”** instead of being identified as the native NEAR asset. This creates incorrect asset representation and may confuse users about the actual token type held in the treasury.

### Preconditions:

- Treasury (Public or Confidential) contains native NEAR balance (https://testenv.trezu.app/staging-qa.sputnik-dao.near)

### Steps To Reproduce

1. Open Treasury Dashboard
2. Navigate to **Assets** section
3. Locate native NEAR asset
4. Observe asset title/description

### Actual Result:

- Native NEAR token is displayed as:

  * `Wrapped NEAR fungible token`

<img width="1365" height="694" alt="Image" src="https://github.com/user-attachments/assets/0b84d732-4692-49cd-9f79-be7513feb879" />

### Expected Result:

- Native NEAR token should be displayed as:

  * `Near Native` (or equivalent native token label)

- Wrapped token naming should only be used for actual wrapped assets (e.g., wNEAR)

### Special Notes:

```
**Possible Root Cause:**

* Asset metadata mapping incorrectly resolves native NEAR to wrapped NEAR token metadata
* UI token registry/conf lookup uses fungible token descriptor for native asset
* Missing differentiation between native chain asset and wrapped FT representation
```

### Workarounds:
None

### Repro Rate:
Always (10/10)

### Setup - OS - Browser - Environment

* OS / Devices: Windows 11 (x64), Motorola G23 (Android 14), Iphone XR (iOs 18.7.6)
* Screen resolution(s): 1366 x 768, 1600 x 720, 1792×828
* Browser(s) + version(s): Chrome (148.0.7778.97), Mozilla Firefox (147.0.2), Brave (1.86.148), Safari (18.7.4), Edge (145.0.3800.65)
* Wallet: Meteor
* App environment: Staging ([trezu.app](https://testenv.trezu.app/confidential-staging.sputnik-dao.near/))
* Near Network: Mainnet

### Impact:

**User Impact:**

* Users may confuse native NEAR with wrapped NEAR token
* Incorrect asset labeling reduces trust in portfolio accuracy

**System Impact:**

* Asset metadata representation is inconsistent
* Native and wrapped asset identities are not properly separated in UI

[...]

PR #733, opened 2026-05-26
fix: normalize NEAR naming and polish network labels/icons

Megha-Dev-19:
- Force NEAR and wrap.near metadata to always use `NEAR` for name/symbol in backend. #676 #724
- Show destination network as `near.com` (instead of expanded label) in transfer details #723
- Make payment network icons round in both recipient network select and send-token network displays #711

### DIFF
--- a/nt-be/data/tokens.json
+++ b/nt-be/data/tokens.json
@@ -355,7 +355,7 @@
     {
       "unifiedAssetId": "near",
       "symbol": "NEAR",
-      "name": "NEAR Protocol",
+      "name": "NEAR",
       "icon": "https://s2.coinmarketcap.com/static/img/coins/128x128/6535.png",
       "groupedTokens": [
         {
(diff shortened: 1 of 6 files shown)


Megha-Dev-19:
@frol @olskik12 quick question: we display wNEAR as NEAR (fungible token) across the app, but in CSV export it currently shows as wNEAR.
How do we want to handle this in export — keep wNEAR for accuracy, or align with UI and show NEAR?

olskik12:
I would suggest using the “NEAR”  token everywhere to keep things consistent, because otherwise it may be confusing why we show one token name in the UI and a different one in the export

ktarun1419 (approved):

Megha-Dev-19:
Note: I’ll look into the failing test in a separate PR. I’ve confirmed it is not related to the current changes.

Issue #756, opened 2026-05-28
[BUG] Trezu - Confidential/Public Payments: Payment amount incorrectly displays `0.000001 NEAR` instead of `1 NEAR`


rknearn-qa:
### Summary

In the **Confidential/Public Treasury Payments** flow, entering a payment amount of `1 NEAR` results in the Review Payment step displaying the amount incorrectly as `0.000001 NEAR`. This indicates a major decimal/precision conversion issue in amount formatting or network/token normalization.

### Preconditions:

- Confidential or Public Treasury exists (https://testenv.trezu.app/staging-qa.sputnik-dao.near/payments)
- Payments flow is accessible

### Steps To Reproduce

1. Open Payments page
2. Select `1 NEAR` on **BNB Chain** network
3. Enter recipient address

   * Example: `0xc366efeff224ea755118b19c30202869c5a85490`
4. Choose **NEAR Public Network**
5. Click **Review Payment**
6. Observe displayed payment amount

### Actual Result:

- Review screen displays:

  * `0.000001 NEAR`

- Entered value `1 NEAR` is converted/displayed incorrectly

<img width="1365" height="683" alt="Image" src="https://github.com/user-attachments/assets/93078939-b042-4898-9c80-d0c0889575ff" />
<img width="1365" height="682" alt="Image" src="https://github.com/user-attachments/assets/f7ebeb67-1a3d-4808-8d30-eb2855344248" />

https://github.com/user-attachments/assets/2a3639c1-4031-4f93-a339-4baa21234181

### Expected Result:

- Review screen should display the correct entered amount:

  * `1 NEAR`

- Amount formatting and token decimal conversion should remain accurate across networks

### Special Notes:

```
**Possible Root Cause:**

* Incorrect token decimal conversion between BNB Chain and NEAR network representations
* Amount normalization uses wrong decimal precision (e.g., 6 decimals instead of NEAR precision)
* Unit conversion logic incorrectly applies smallest-unit formatting during review stage
* Cross-network asset mapping incorrectly transforms displayed value

```

### Workarounds:
None

### Repro Rate:
Always (10/10)

### Setup - OS - Browser - Environment

* OS / Devices: Windows 11 (x64), Motorola G23 (Android 14), Iphone XR (iOs 18.7.6)
* Screen resolution(s): 1366 x 768, 1600 x 720, 1792×828
[...]

PR #771, opened 2026-06-01
fix: use destination network decimals for EXACT_OUTPUT quote amounts

Megha-Dev-19:
#756 
We used decimals from the selected token, which gives wrong amounts when the destination token uses a different decimal count, fixed quote amount scaling for `EXACT_OUTPUT` payments by using decimals from the destination network token.

### DIFF
--- a/nt-fe/hooks/use-intents-quote.ts
+++ b/nt-fe/hooks/use-intents-quote.ts
@@ -4,6 +4,7 @@ import { useCallback, useMemo, useState } from "react";
 import { useDebounce } from "use-debounce";
 import { useTranslations } from "next-intl";
 import { useQuery } from "@tanstack/react-query";
+import * as Sentry from "@sentry/nextjs";
 import { NEAR_NETWORK_ID } from "@/constants/network-ids";
 import { getAddressPattern } from "@/lib/address-validation";
 import Big from "@/lib/big";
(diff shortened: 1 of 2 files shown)


nt-fe/hooks/use-intents-quote.ts
```diff
@@ -175,14 +177,17 @@ export function useIntentsQuote({
     const [debouncedAddress] = useDebounce(address, 300);
     const [debouncedAmount] = useDebounce(amount, 400);
     const [isEnsuring, setIsEnsuring] = useState(false);
+    const effectiveAmountDecimals = destinationAmountDecimals ?? token.decimals;
```
frol: Is it really a good fallback?

frol (approved):

nt-fe/hooks/use-intents-quote.ts
Megha-Dev-19: For most tokens, decimals are the same across networks.
A few tokens have different decimals on different networks.

I tested those mixed-decimal tokens, and the new logic works correctly.

If destination metadata is missing, I intentionally fall back to sender token decimals. This is not perfect, but it prevents the flow from breaking completely.

In that fallback case, users should still notice something is wrong in review (token amount / USD will look off before submit).

Do you think we should throw an error in the fallback case instead?

frol: Yeah, it is better to retry and fail if it doesn’t improve (display an error and report through Sentry - I guess you just need to raise an exception)

Issue #786, opened 2026-06-03
[BUG] Trezu - Exchange: Generated PDF receipt displays astronomically incorrect received amount and exchange rate


rknearn-qa:
### Summary

The **Generate PDF Receipt** functionality displays severely incorrect values for the received NEAR amount and exchange rate. Instead of showing the actual transaction amount, the receipt contains values in the range of **10²²–10²⁴ NEAR**, indicating a likely token decimal conversion or yoctoNEAR normalization failure.

The transaction itself is correct in Trezu, but the generated receipt presents invalid financial information.

### Preconditions:

- Exchange request has been executed successfully (https://testenv.trezu.app/staging-qa.sputnik-dao.near/requests/79)

### Steps To Reproduce

1. Open completed Exchange request
2. Click **Generate PDF Receipt**
3. Open generated receipt
4. Compare receipt values with actual transaction details

Receipt URL:

```
https://testenv.trezu.app/staging-qa.sputnik-dao.near/requests/79/receipt
```

### Actual Result:

- Receipt displays incorrect financial values:

```Received Amount:
10,000,000,000,000,000,000,000.00 NEAR
```

```
Rate:
1 NEAR ($2.60)
≈ 1,000,000,000,000,000,000,000,000.00 NEAR
```

- Values exceed actual transaction amount by many orders of magnitude
- Financial information in the receipt is unusable and misleading

<img width="1365" height="684" alt="Image" src="https://github.com/user-attachments/assets/428e504a-5d00-41dc-b063-7f014331f47d" />

### Expected Result:

Receipt should display the actual transaction values:

```
Received Amount:
0.01 NEAR
```

```
Rate:
1 NEAR ($2.95)
≈ 1.000000 NEAR
```

- Receipt values should match request details and on-chain transaction data
- Token amounts should be correctly normalized and formatted

### Special Notes:

```
**Possible Root Cause:**

* yoctoNEAR values are rendered without conversion to NEAR units
* Decimal normalization is applied incorrectly during PDF generation
* Receipt formatter uses raw blockchain values instead of display values
* Token precision metadata is missing or ignored in receipt template
* Currency conversion calculation is performed against smallest-unit values
```

### Workarounds:

- Verify amounts directly in Trezu Request Details instead of relying on generated PDF receipts

### Repro Rate:

- Always (10/10)

### Setup - OS - Browser - Environment

[...]

PR #791, opened 2026-06-03
fix: receipt wrap/unwrap amounts shown as raw yoctoNEAR

Megha-Dev-19:
#786


frol (commented):
@Megha-Dev-19 Is `amountOut` invalid only for wrap.near? Do we know why? Can we fix the root cause?

Megha-Dev-19:
@frol this happens because it is not an intents exchange request, it is an `ft_transfer` to `wrap.near` to convert native NEAR to FT.

I moved the extraction to the top level where the request is parsed, so all files receive in same way.

frol (approved):

Issue #781, opened 2026-06-02
[DEV] Improve the creation flow and user onboarding (stepper in the dashboard) 2.0


olskik12:
Implement #738 

Acceptance Criteria

- [ ] When a user navigates to Trezu from the landing page, they should see the treasury creation flow.

<img width="740" height="473" alt="Image" src="https://github.com/user-attachments/assets/44440668-0ae9-4938-9462-f6541867f7e2" />

- [ ] Connect walet screen for first time users

<img width="674" height="425" alt="Image" src="https://github.com/user-attachments/assets/da12a69c-c64e-4cc8-a0ab-e176365f80de" />

- [ ] Users who already have a treasury should see a “Sign In” button at the bottom of the page.
- [ ] For existing users, the wallet connection flow should display the wallet they most recently connected with.

<img width="698" height="436" alt="Image" src="https://github.com/user-attachments/assets/c8a9d27d-0db5-4990-865f-379ee8d40f39" />

- [ ] On the Dashboard, users should see three onboarding steps:

- Add a team member
- Add your first asset
- Create your first payment request

- [ ] If a user chooses to add a member from the Dashboard page, they should be redirected to the Members page.

[Figma](https://www.figma.com/design/UlUkb3wL7fQq0xNvmjFnEg/TREZU?node-id=8019-8037&t=JqnHYrSS8yeWI9u5-4) | [Prototype](https://www.figma.com/proto/UlUkb3wL7fQq0xNvmjFnEg/TREZU?node-id=8093-14453&viewport=219%2C-1318%2C0.07&t=xEr9bokM8STIlRhq-8&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=8093%3A14453&page-id=6182%3A62543&hide-ui=1)

rknearn-qa:
Implemented

<img width="1365" height="684" alt="Image" src="https://github.com/user-attachments/assets/5b5dbb51-cfdd-44f9-88ff-206ec62f4c55" />

* OS / Devices: Windows 11 (x64), Motorola G23 (Android 14), Iphone XR (iOs 18.7.6)
* Screen resolution(s): 1366 x 768, 1600 x 720, 1792×828
* Browser(s) + version(s): Chrome (149.0.7827.54), Mozilla Firefox (151.0.3), Brave (1.91.168), Safari (18.7.4), Edge (149.0.4022.52)
* Wallet: Meteor
[...]

PR #829, opened 2026-06-08
feat: update onboarding create flow and add members setup guidance

Megha-Dev-19:
#781

image
image
image

image
image


frol:
> Note: For MetaMask I use walletcontract-eip712 wallet id, and for Binance + WalletConnect I use wallet-connect.
If MetaMask is installed, selecting Binance/WalletConnect can otherwise open MetaMask extension directly and not let user choose anything else. Using wallet-connect for those keeps it on QR flow so user can scan from Binance app.

image

1. I suggest we don't re-invent the wheel and let WalletConnect handle all those wallets - I believe Ethereum users are used to it as-is, so we don't need to split it into our own list of wallets
2. `wallet-connect` won't work for Binance and most other WalletConnect wallets. It is the near-compatible integration with WalletConnect, which only works for UnityWallet, but I don't know any other EVM wallet that supports it. Thus, you must use `walletcontract-eip712`

frol:
Also, please, test that after the changes to NEAR Connect, it does not log users out on every browser page reload

Megha-Dev-19:
@frol oh, I didn’t realize the WalletConnect option was NEAR-compatible, I saw it in the manifest and assumed it would work for all wallets. Thanks for pointing that out.
Now, when a user selects an EVM wallet, we directly open NEAR Connect with `walletcontract-eip712`.
I also verified that login status persists correctly after refresh.

nt-fe/messages/es.json
```diff
@@ -778,6 +777,18 @@
             "removeMember": "Actualizar política - Eliminar miembro",
             "removeMembers": "Actualizar política - Eliminar miembros",
             "removeMemberSuccess": "Solicitud de eliminación de miembro creada correctamente"
+        },
+        "infoSection": {
```
frol, line 781: For some reason this section of translations is not translated for all the languages

frol (approved):
Fix the translations, and feel free to merge it

Issue #828, opened 2026-06-08
[BUG] Trezu - PDF Receipt: Small token amounts are rounded to $0.00 in generated receipt


rknearn-qa:
### Summary

Generated PDF receipts incorrectly display the USD value as **$0.00** for low-value token amounts, even when the transaction has a non-zero USD equivalent.

This results in inaccurate financial reporting within the receipt and may cause confusion during accounting, auditing, or transaction verification.

### Preconditions:

- Payment request has been executed
- Payment amount is small enough to produce a USD value below the current display precision

### Steps To Reproduce

1. Execute a payment with a small token amount

   * Example: `0.0001 NEAR`
2. Open Request Details
3. Generate a PDF Receipt (https://testenv.trezu.app/confidential-staging.sputnik-dao.near/requests/83/receipt)
4. Observe the Amount (USD) section

### Actual Result:

PDF receipt displays:

<img width="1365" height="690" alt="Image" src="https://github.com/user-attachments/assets/329d78dd-b2e2-47a4-bf8c-47b35d12ddf4" />

USD value is shown as zero despite the transaction having a non-zero fiat value

### Expected Result:

- PDF receipt should display the actual calculated USD value
- Small values should retain sufficient precision

Examples:

```text
$0.0003
```

or

```text
<$0.01
```

or another agreed formatting approach that preserves the fact that the value is non-zero.

### Special Notes:

```
**Possible Root Cause:**

* USD values are rounded to two decimal places before rendering
* Values below `$0.01` are truncated to `$0.00`
* Receipt formatting logic does not support low-value asset transactions
* PDF generation uses display formatting intended for dashboard views rather than accounting records
```

### Workarounds:

- Refer to transaction details outside the generated PDF

### Repro Rate:
Always (10/10)

### Setup - OS - Browser - Environment

* OS / Devices: Windows 11 (x64), Motorola G23 (Android 14), Iphone XR (iOs 18.7.6)
* Screen resolution(s): 1366 x 768, 1600 x 720, 1792×828
* Browser(s) + version(s): Chrome (149.0.7827.54), Mozilla Firefox (151.0.3), Brave (1.91.168), Safari (18.7.4), Edge (149.0.4022.52)
* Wallet: Meteor
* App environment: Staging ([trezu.app](https://testenv.trezu.app/confidential-staging.sputnik-dao.near/requests?tab=Approved))
* Near Network: Mainnet

### Impact:

**User Impact:**

[...]

Issue #604, opened 2026-05-06
[BUG] Trezu - Dashboard: BLACKDRAGON token displays incorrect price ($0.00)


rknearn-qa:
### Summary

In the **Dashboard**, the **BLACKDRAGON token** displays an incorrect **price of $0.00**, despite having a non-zero total valuation. This results in inconsistent asset valuation and misleading portfolio data.

### Preconditions:

- Treasury contains **BLACKDRAGON token** (https://testenv.trezu.app/staging-qa.sputnik-dao.near)
- Example state:

  * Balance: **301,503,941.3402 BLACKDRAGON**
  * Total value: **~$1.80**

### Steps To Reproduce

1. Open **Dashboard**
2. Locate **BLACKDRAGON token** in asset list
3. Observe **token price and total value**

### Actual Result:

- Token price displays as: **$0.00**
- Total value is non-zero (e.g., ~$1.80)
- Price and total value are inconsistent

<img width="1365" height="689" alt="Image" src="https://github.com/user-attachments/assets/2e8bb5d9-2227-49f8-8521-8fa3e1d6495a" />

### Expected Result:

- Token price should reflect **actual market price**
- Total value should be calculated correctly based on price × balance
- No inconsistency between displayed price and total value

### Special Notes:

```
Possible Root Cause

* Price feed returns **very small value**, rounded to 0.00 in UI
* Incorrect decimal handling or rounding logic
* Price API missing precision or returning null/zero fallback
* UI formatting truncates small values instead of displaying scientific/extended precision

```

### Workarounds:
None

### Repro Rate:
Always (10/10)

### Setup - OS - Browser - Environment

* OS / Devices: Windows 11 (x64), Motorola G23 (Android 14), Iphone XR (iOs 18.7.6)
* Screen resolution(s): 1366 x 768, 1600 x 720, 1792×828
* Browser(s) + version(s): Chrome (148.0.7778.97), Mozilla Firefox (147.0.2), Brave (1.86.148), Safari (18.7.4), Edge (145.0.3800.65)
* Wallet: Meteor
* App environment: Staging (trezu.app)
* Near Network: Mainnet

### Impact:

**User Impact:**

* Misleading asset pricing information
* Users cannot accurately assess token value

**System Impact:**

* Inconsistent pricing logic between unit price and total value
* Potential issues with price formatting or precision handling

**Business / Release Risk:**

* Medium risk — affects trust in portfolio accuracy
* Impacts perception of supported token reliability

[...]

PR #832, opened 2026-06-09
feat: show <$0.01 for sub-cent USD values

Megha-Dev-19:
#828 #604 

image
image

### DIFF
--- a/nt-fe/lib/utils.ts
+++ b/nt-fe/lib/utils.ts
@@ -29,12 +29,38 @@ export function base64ToJson(base64: string): any {
     return JSON.parse(decoded);
 }
 
-export function formatCurrency(value: number | Big) {
+interface FormatCurrencyOptions {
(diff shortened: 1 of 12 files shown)


nt-fe/lib/utils.ts
```diff
@@ -29,12 +29,38 @@ export function base64ToJson(base64: string): any {
...
+    const { showSubCentAsLessThan = false, minimumFractionDigits = 2 } =
+        options;
+
+    if (
+        showSubCentAsLessThan &&
+        absoluteValue.gt(0) &&
+        absoluteValue.lt(0.01)
+    ) {
+        return bigValue.lt(0) ? "-<$0.01" : "<$0.01";
+    }
```
ktarun1419, line 61: we can add a finite value check as well
`NaN`, `Infinity`, values can break

ktarun1419 (approved):

PR #836, opened 2026-06-09
feat: added tg notification for sponsor low balance

ktarun1419:
Added a worker to check the sponsor balance every 1 minute and alert if the sponsor balance is lower than the threshold

### DIFF
--- a/nt-be/src/services/sponsor_alerts.rs
+++ b/nt-be/src/services/sponsor_alerts.rs
@@ -1,0 +1,155 @@
+use std::sync::{Arc, LazyLock};
+use std::time::{Duration, Instant};
+
+use near_api::{NearToken, Tokens};
+use tokio::sync::RwLock;
+
+use crate::{AppState, constants::LOW_BALANCE_THRESHOLD, utils::telegram::TelegramClient};
+
+const ALERT_COOLDOWN: Duration = Duration::from_secs(60 * 60);
+const DEFAULT_POLL_INTERVAL_SECS: u64 = 60;
+const INITIAL_DELAY_SECS: u64 = 20;
+
+static LAST_ALERT_SENT_AT: LazyLock<RwLock<Option<Instant>>> = LazyLock::new(|| RwLock::new(None));
+
+/// Returns true when liquid balance is below the sponsor low-balance threshold.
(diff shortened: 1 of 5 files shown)


nt-be/src/services/sponsor_alerts.rs
```diff
@@ -0,0 +1,155 @@
...
+    }
+}
+
+pub(crate) fn format_near_display(amount: NearToken) -> String {
+    const ONE_NEAR: u128 = 10u128.pow(24);
+    let yocto = amount.as_yoctonear();
+    let whole = yocto / ONE_NEAR;
+    let frac = (yocto % ONE_NEAR) * 10_000 / ONE_NEAR;
+    format!("{}.{:04}", whole, frac)
+}
```
frol: NearToken already has Display implemented, so no need for a custom formatter.

nt-be/src/services/sponsor_alerts.rs
```diff
@@ -0,0 +1,155 @@
...
+use std::time::{Duration, Instant};
+
+use near_api::{NearToken, Tokens};
+use tokio::sync::RwLock;
+
+use crate::{AppState, constants::LOW_BALANCE_THRESHOLD, utils::telegram::TelegramClient};
+
+const ALERT_COOLDOWN: Duration = Duration::from_secs(60 * 60);
+const DEFAULT_POLL_INTERVAL_SECS: u64 = 60;
+const INITIAL_DELAY_SECS: u64 = 20;
```
frol: Use const `Duration::from_secs` instead of integer consts.

nt-be/src/services/sponsor_alerts.rs
```diff
@@ -0,0 +1,155 @@
...
+
+    if !is_balance_low(liquid) {
+        return Ok(());
+    }
+
+    let now = Instant::now();
+    let last_sent = *LAST_ALERT_SENT_AT.read().await;
+    if !cooldown_allows_alert(last_sent, now) {
+        return Ok(());
+    }
```
frol, line 59: There is no point in querying the balance from RPC if we don't plan to alert on it, so I suggest to move this check before the fetch call.

Megha-Dev-19 (approved):

PR #840, opened 2026-06-09
feat: show waitlist on treasury creation failure

Megha-Dev-19:
#846 
Now we show the waitlist if any type of error occurred during creation, and log the error in the backend.
image
image


ktarun1419 (approved):

frol (approved):

olskik12:
Hi @Megha-Dev-19 Could you please confirm the check icon size on the success screen? In Figma, we have an icon component sized 32×32 px (you can refer to the request page). Could you please use the same size here as well?

Issue #464, opened 2026-04-07
[BUG] Trezu - Address Book: Xlayer network displayed with typo as "LayerX"


rknearn-qa:
### Summary

In the **Address Book module**, the **Xlayer network** is incorrectly displayed as **"LayerX"**, which is a typo. This leads to inconsistency with the correct network naming used across the system and may confuse users selecting networks.

### Preconditions:

- User has a member role for this Treasury (https://testenv.trezu.app/romakqatesting.sputnik-dao.near)
- Address Book module is accessible

### Steps To Reproduce

1. Navigate to **Address Book**
2. Open network selection (e.g., while adding an address)
3. Locate **Xlayer** network in the list

### Actual Result:

- Network is displayed as **"LayerX"**

<img width="1365" height="685" alt="Image" src="https://github.com/user-attachments/assets/a82af358-f0b1-49ec-bd3c-3567d0138080" />

<img width="1365" height="686" alt="Image" src="https://github.com/user-attachments/assets/4a7c533a-2894-4dd6-a2f9-d2a16e3d10c2" />

### Expected Result:

- Network should be displayed as **"Xlayer"** consistently across the application

### Special Notes:

```
Possible Root Cause

* Likely a **hardcoded typo** or incorrect label mapping in the network configuration or localization file (e.g., enum/key mismatch such as `LAYER_X` vs `XLAYER`)
```

### Workarounds:
None

### Repro Rate:
Always

### Setup - OS - Browser - Environment

* OS / Devices: Windows 11 (x64), Motorola G23 (Android 14), Iphone XR (iOs 18.7.6)
* Screen resolution(s): 1366 x 768, 1600 x 720, 1792×828
* Browser(s) + version(s): Chrome (146.0.7680.178), Mozilla Firefox (147.0.2), Brave (1.86.148), Safari (18.7.4), Edge (145.0.3800.65)
* Wallet: Meteor
* App environment: Staging ([trezu.app](https://testenv.trezu.app/romakqatesting.sputnik-dao.near))
* Near Network: Mainnet

### Impact:

**User Impact:**

* Users may be confused by inconsistent or incorrect network naming
* Reduces trust in UI accuracy

**System Impact:**

* Indicates inconsistency in network configuration or localization

**Business / Release Risk:**

* Minor UX issue, but affects product polish and professionalism

### QA Verdict:

**Verdict:** ❌ Minor Bug
**Rationale:**

[...]

Issue #433, opened 2026-03-30
[BUG] Trezu - Deposit: NEAR network icon not displayed for “Other” asset type (fallback icon shown)


rknearn-qa:
### Summary

In the **Deposit** module, when selecting **“Other” asset type**, the **NEAR network icon is not displayed correctly**. Instead of the custom NEAR SVG icon, a **default “N” placeholder icon** is shown, indicating missing or incorrect asset mapping.

### Preconditions:

- User has **Requestor role** in Treasury

### Steps To Reproduce

1. Navigate to **Deposit**
2. Select **“Other”** asset type
3. Observe the **NEAR network icon**

### Actual Result:

- Default placeholder icon (“N”) is displayed
- Custom NEAR SVG icon is not applied

<img width="1365" height="683" alt="Image" src="https://github.com/user-attachments/assets/43c84921-d9bd-4519-9c79-edb6567d1675" />

### Expected Result:

- Correct NEAR SVG icon should be displayed
- Icon should be consistent with other modules (e.g., Address Book, Payments, Deposit NEAR FT/Native/Intents)

<img width="1365" height="689" alt="Image" src="https://github.com/user-attachments/assets/905f6264-a9be-444a-b918-4a5b7a25c7aa" />

### Special Notes:

```
Possible Root Cause

* Missing icon mapping for NEAR under “Other” asset type
* Incorrect asset-type-based icon resolution logic
```

### Workarounds:
None

### Repro Rate:
Always

### Setup - OS - Browser - Environment

* OS / Devices: Windows 11 (x64), Motorola G23 (Android 14), Iphone XR (iOs 18.7.6)
* Screen resolution(s): 1366 x 768, 1600 x 720, 1792×828
* Browser(s) + version(s): Chrome (144.0.7559.97), Mozilla Firefox (147.0.2), Brave (1.86.148), Safari (18.7.4), Edge (145.0.3800.65)
* Wallet: Meteor
* App environment: Staging ([trezu.app](https://testenv.trezu.app/romakqatesting.sputnik-dao.near))
* Near Network: Mainnet

### Impact:

**User Impact:**

* Minor visual inconsistency
* Reduced clarity in identifying network

**System Impact:**

* Indicates missing asset mapping for specific asset type

**Business / Release Risk:**

* Low, but affects UI consistency and polish

### QA Verdict:

**Verdict:** Minor Issue
**Rationale:**

* Visual inconsistency in a key flow (Deposit)
[...]

PR #851, opened 2026-06-10
Fix/near icon in depsoit modal

ktarun1419:
#433 #464


Megha-Dev-19 (approved):

Megha-Dev-19:
@ktarun1419 didn't we had other ticket for layerX?
can you attach that in here?

Issue #759, opened 2026-05-29
[BUG] Trezu - Bulk Payment: Validation message is visually cropped


rknearn-qa:
### Summary

In the **Bulk Payment** flow, validation messages displayed after entering invalid data are visually cropped/truncated in the UI. Users cannot read the full validation text, making it difficult to understand how to correct formatting issues in uploaded/pasted payment data.

### Preconditions:

- Bulk Payment flow is accessible

### Steps To Reproduce

1. Open **Bulk Payment**
2. Navigate to **Provide Data** step
3. Enter invalid address/data

   * Example: `amount.near`
4. Click **Continue to Review**
5. Observe validation message

### Actual Result:

- Validation message is cropped/truncated in UI
- Full message is not readable

<img width="1365" height="688" alt="Image" src="https://github.com/user-attachments/assets/97db31d4-b372-44fd-92c1-c2db2157572a" />

### Expected Result:

- Validation message should display fully without cropping
- Text container should support multiline wrapping and responsive sizing
- Users should be able to read the complete validation guidance

### Special Notes:

```
**Possible Root Cause:**

* Validation container has fixed height/overflow hidden styling
* Text wrapping is disabled or constrained by parent container width
* Responsive layout does not handle long validation strings correctly
* Error tooltip/banner component lacks multiline support
```

### Workarounds:
None

### Repro Rate:
Always (10/10)

### Setup - OS - Browser - Environment

* OS / Devices: Windows 11 (x64), Motorola G23 (Android 14), Iphone XR (iOs 18.7.6)
* Screen resolution(s): 1366 x 768, 1600 x 720, 1792×828
* Browser(s) + version(s): Chrome (148.0.7778.97), Mozilla Firefox (147.0.2), Brave (1.86.148), Safari (18.7.4), Edge (145.0.3800.65)
* Wallet: Meteor
* App environment: Staging ([trezu.app](https://testenv.trezu.app/paul-morphy-test.sputnik-dao.near/payments/bulk-payment))
* Near Network: Mainnet

### Impact:

**User Impact:**

* Users cannot fully understand validation requirements
* Increases confusion when formatting bulk payment data

**System Impact:**

* Validation/error presentation is visually broken
* UI responsiveness/accessibility is reduced

**Business / Release Risk:**

* Low risk — UI/UX issue only
* May increase onboarding/support friction for Bulk Payment users

### QA Verdict:

**Verdict:** ❌ Needs Fix
**Rationale:**

[...]

Issue #404, opened 2026-03-24
[BUG] Trezu - Address Book: Deleted address remains visible across sessions (no data sync)


rknearn-qa:
### Summary

In the **Address Book** module, when a recipient is deleted in one browser/session, the same address **remains visible in another browser/tab/window**. The data is not automatically refreshed, leading to stale state. Users can attempt to delete the already removed address again, indicating lack of **real-time sync or data re-fetching**.

### Preconditions:

- At least one address is created in Address Book
- Application is opened in multiple browsers/tabs/windows (different sessions)

### Steps To Reproduce

1. Open Address Book in Browser A
2. Open Address Book in Browser B (same account)
3. In Browser A, delete a recipient
4. Switch to Browser B
5. Attempt to delete the same recipient

### Actual Result:

- Deleted address is still visible in Browser B
- User can attempt to delete the same address again
- Data is not automatically updated

https://github.com/user-attachments/assets/7d0d941a-9281-4fa4-807f-dee04455d4bf

<img width="1365" height="680" alt="Image" src="https://github.com/user-attachments/assets/acf6cc49-6162-437f-b182-2645957819e9" />

### Expected Result:

- Address Book should reflect the latest data across sessions
- Deleted address should no longer be visible in other tabs/browsers
- Data should be auto-refetched or synchronized (e.g., via polling, cache invalidation, or real-time updates)

### Special Notes:

[Design](https://www.figma.com/design/UlUkb3wL7fQq0xNvmjFnEg/TREZU?node-id=4919-126312&t=5oyB9mkw3I4v0zph-0)

```
Possible Root Cause

* Missing cache invalidation or stale client-side state (e.g., React Query, Redux store not updated)
* No background refetch/polling or WebSocket subscription for data updates
* API response not revalidated on focus or action
* Lack of synchronization between sessions
```

### Workarounds:

- Manually refresh the page to fetch updated data

### Repro Rate:
Always

### Setup - OS - Browser - Environment

* OS / Devices: Windows 11 (x64), Motorola G23 (Android 14), Iphone XR (iOs 18.7.6)
* Screen resolution(s): 1366 x 768, 1600 x 720, 1792×828
* Browser(s) + version(s): Chrome (144.0.7559.97), Mozilla Firefox (147.0.2), Brave (1.86.148), Safari (18.7.4), Edge (145.0.3800.65)
* Wallet: Meteor
[...]

PR #860, opened 2026-06-11
chore: UI improvements

ktarun1419:
#759 
 partially: #404

### DIFF
--- a/nt-fe/features/address-book/hooks/use-address-book.ts
+++ b/nt-fe/features/address-book/hooks/use-address-book.ts
@@ -12,8 +12,9 @@ export function useAddressBook() {
         queryKey: ["address-book", treasuryId, accountId],
         queryFn: () => getAddressBook(treasuryId!),
         enabled,
-        staleTime: 1000 * 60, //1 minute
+        staleTime: 1000* 10, //10 seconds
         refetchInterval: 1000 * 60, //1 minute
         refetchIntervalInBackground: true,
+        refetchOnWindowFocus: true
     });
 }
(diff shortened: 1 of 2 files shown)


nt-fe/features/address-book/hooks/use-address-book.ts
```diff
@@ -12,8 +12,9 @@ export function useAddressBook() {
         queryKey: ["address-book", treasuryId, accountId],
         queryFn: () => getAddressBook(treasuryId!),
         enabled,
-        staleTime: 1000 * 60, //1 minute
+        staleTime: 1000 * 10, //10 seconds
         refetchInterval: 1000 * 60, //1 minute
         refetchIntervalInBackground: true,
+        refetchOnWindowFocus: true,
```
Megha-Dev-19, line 18: I don’t think we should refetch every 10 seconds, that’s too many calls.  
1 minute should be fine, since users usually take about a minute to add/remove entries.

Also, having the page open in two browsers at the same time feels like an edge case. In practice, address book is usually managed by one team member.

Megha-Dev-19 (approved):

nt-fe/features/address-book/hooks/use-address-book.ts
ktarun1419: Actually, the refetch interval is still 1 minute, not 10 seconds. The `staleTime: 10s` only means the cached data becomes stale after 10 seconds.

So we are not polling every 10 seconds. It will refetch every 1 minute, and also refetch on window focus only if the data is already stale. I think this is fine for keeping the address book reasonably fresh without too many calls.

Issue #528, opened 2026-04-22
[BUG] Trezu - Onboarding: “Create a Payment Request” banner does not update after request completion


rknearn-qa:
### Summary

In the **Onboarding flow**, the banner step **“3. Create a Payment Request”** remains **incomplete and visible** even after a payment request is **created and approved**. The onboarding state is not updated, and banners do not disappear as expected.

### Preconditions:

- Treasury is **newly created**
- First two onboarding steps are completed

### Steps To Reproduce

1. Create a **Payment request**
2. Approve the payment request
3. Return to the **Dashboard / Onboarding section**

### Actual Result:

- Onboarding banners are still displayed
- Step **“3. Create a Payment Request”** remains **incomplete**

<img width="1365" height="690" alt="Image" src="https://github.com/user-attachments/assets/7303f192-708a-4719-b7f9-50ae54cca769" />
<img width="1365" height="683" alt="Image" src="https://github.com/user-attachments/assets/e0f921f4-7074-4ed4-b6f2-d0c8685c7b52" />
<img width="1365" height="684" alt="Image" src="https://github.com/user-attachments/assets/11cd8bc9-77e7-4c07-b3ef-44675047f4c1" />

### Expected Result:

- Step **“3. Create a Payment Request”** should be marked as **completed**
- All onboarding banners should **disappear** after completion

Approved [Design](https://www.figma.com/design/UlUkb3wL7fQq0xNvmjFnEg/TREZU?node-id=598-16866&t=pHL8AnNFXUyCTbAu-4)

### Special Notes:

```
Possible Root Cause

* Missing update of **onboarding progress state** after request approval
* Event for **payment creation/approval** is not linked to onboarding logic
* Frontend may rely on stale state or missing refetch after action
```

### Workarounds:
None

### Repro Rate:
Always (10/10)

### Setup - OS - Browser - Environment

* OS / Devices: Windows 11 (x64), Motorola G23 (Android 14), Iphone XR (iOs 18.7.6)
* Screen resolution(s): 1366 x 768, 1600 x 720, 1792×828
* Browser(s) + version(s): Chrome (146.0.7680.178), Mozilla Firefox (147.0.2), Brave (1.86.148), Safari (18.7.4), Edge (145.0.3800.65)
* Wallet: Meteor
* App environment: Staging (https://testenv.trezu.app/confidential-staging.sputnik-dao.near)
* Near Network: Mainnet

### Impact:

**User Impact:**

[...]

PR #861, opened 2026-06-11
fix: fixed payment step marking issue

ktarun1419:
#528 
The main issue was a race condition. We fetch the proposals 2 seconds after submitting one, but our indexer may not have processed it by then.
  
  - Added a cache key to track when a proposal has been submitted
  - Selectively refetches every 5 seconds when a cache key is detected until the proposal is returned, or for a maximum of 3 minutes


Megha-Dev-19:
@ktarun1419 from my manual testing, proposals always show up in the list after I create one.
User flow is usually: create proposal from Payment, they get toast notification that it's success, then they can go to Dashboard/Requests.  
If we always refetch proposals on Dashboard load, shouldn’t it always appear?
I m not sure if we need the polling, we can do `refetchOnMount` and that should work ?

ktarun1419:
> @ktarun1419 from my manual testing, proposals always show up in the list after I create one.
> User flow is usually: create proposal from Payment, they get toast notification that it's success, then they can go to Dashboard/Requests.
> If we always refetch proposals on Dashboard load, shouldn’t it always appear?
> I m not sure if we need the polling, we can do `refetchOnMount` and that should work ?

Yes, refetchOnMount will help when the user opens the Dashboard.

But the success toast only confirms that the on-chain transaction is done. It does not confirm that our backend indexer has picked up the new proposal yet.

The proposals list comes from the indexer. So if the user opens Dashboard before the indexer has updated, one refetchOnMount call can still miss it.

That is why polling is useful here. It keeps retrying for a short time until the proposal appears, and then it stops. So I think we should keep both

nt-fe/features/onboarding/steps/dashboard.tsx
```diff
@@ -342,8 +346,19 @@ export function CongratsTooltip() {
...
+            refetchInterval: buildPaymentPendingRefetchInterval(treasuryId),
+        },
     );
 
+    useEffect(() => {
+        if (treasuryId && (proposals?.proposals?.length ?? 0) > 0) {
+            clearPaymentPending(treasuryId);
+        }
+    }, [proposals, treasuryId]);
+
```
Megha-Dev-19: why do we want to do it in dashboard also?
isn't doing it in onboarding-progress sufficient?

ktarun1419: Both are always co-mounted on the same page, so the poll started by OnboardingProgress already refreshes the data

Good call.

Megha-Dev-19 (approved):

Issue #720, opened 2026-05-23
[BUG] Deposit BTC on NEAR gives incorrect deposit address


frol:
### Summary

<img width="515" height="646" alt="Image" src="https://github.com/user-attachments/assets/5f91764c-8e27-4063-8c15-7e569385b869" />

### Steps To Reproduce

Open deposit dialog, select BTC asset, NEAR network.

The expected deposit address on public treasury is xxx.sputnik-dao.near and on confidential treasury is an implicit account name on NEAR (64 hex characters).

### Setup - OS - Browser - Environment

any

### Priority

🔴 P0 : Critical

Megha-Dev-19:
hey @frol
I looked into this issue, and we haven’t changed anything in this part.

Current logic is:
- we show the treasury address only when `chainId` contains `near`, eg: `near:mainnet` [here](https://github.com/NEAR-DevHub/trezu/blob/main/nt-fe/app/(treasury)/%5BtreasuryId%5D/dashboard/components/deposit-modal.tsx#L624)
- for BTC, the `chainId` is `btc:mainnet`, so that condition doesn’t match
- in that case, we use the deposit address returned by Intents

Do we want to update this condition so that when the `chainId` contains `:mainnet`, we show the treasury account instead?

Megha-Dev-19:
The issue is mainly because our token list includes both legacy and new BTC bridge entries. Vlad and I discussed removing the legacy one from frontend.

 Also add:
- `near` network for all public tokens
- `near.com` for all confidential tokens

which will show the treasury account ID instead of the 1Click deposit address.

rknearn-qa:
Retested
Fixed

<img width="1365" height="691" alt="Image" src="https://github.com/user-attachments/assets/e74e0d6a-2ba7-46f3-a43b-ab3a87f1e5f1" />
<img width="1365" height="682" alt="Image" src="https://github.com/user-attachments/assets/7f787399-1bd2-4a6d-afc1-54834eda7a92" />

* OS / Devices: Windows 11 (x64), Motorola G23 (Android 14), Iphone XR (iOs 18.7.6)
* Screen resolution(s): 1366 x 768, 1600 x 720, 1792×828
* Browser(s) + version(s): Chrome (149.0.7827.54), Mozilla Firefox (151.0.3), Brave (1.91.168), Safari (18.7.4), Edge (149.0.4022.52)
* Wallet: Meteor
* App environment: Staging (trezu.app)
* Near Network: Mainnet

PR #874, opened 2026-06-12
feat: add canonical near network in deposit flow

Megha-Dev-19:
#720 

Adds NEAR network handling in deposit flow for public and confidential treasuries.

- Public: NEAR network always resolves to treasury account.
image

- Confidential:
  - Existing NEAR-network assets keep current source behavior.
  - Fallback NEAR entries show only near.com path under NEAR network.
  
image

### DIFF
--- a/nt-be/src/handlers/intents/bridge_tokens.rs
+++ b/nt-be/src/handlers/intents/bridge_tokens.rs
@@ -2,7 +2,11 @@ use crate::{
     constants::intents_tokens::{find_unified_asset_id, get_tokens_map},
     utils::cache::CacheTier,
 };
-use axum::{Json, extract::State, http::StatusCode};
+use axum::{
+    Json,
+    extract::{Query, State},
+    http::StatusCode,
+};
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
 use std::collections::HashMap;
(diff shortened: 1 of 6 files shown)


nt-be/src/handlers/intents/bridge_tokens.rs
```diff
@@ -114,6 +134,10 @@ pub async fn get_bridge_tokens(
                     continue;
                 };
 
+                if BLOCKED_NETWORK_IDS.contains(&intents_id) {
+                    continue;
+                }
```
frol, line 139: Will it block users who already own this token from payments and swaps? Will the tokens still displayed on the dashboard? Can you check if there are any users who already own this token?

near.com has a banner for the users to swap the assets to the new BTC bridge tokens

frol (approved):
Besides the mentioned concern, this PR looks good to me, so if we don't block users by this change, or there are no existing users who own the token, we can merge it

nt-be/src/handlers/intents/bridge_tokens.rs
Megha-Dev-19: We have different endpoint to get the treasury owned assets, so it won't impact that
But I still manually tested with one of our client's treasury which owns it, and it works fine
The icon looks different in my local, as I have a different stored one in my db so please ignore that.

image
image

Issue #880, opened 2026-06-12
[BUG] Payment recipient network is auto-selected incorrectly


frol:
### Summary

* Avax network is automatically selected on any recipient address
* near.com network is automatically selected even when there is NEAR network is also available (so it is never one option and as such the user must manually pick the network)

### Steps To Reproduce

Open the payment screen and enter the recipient address for USDC token.

<img width="1362" height="690" alt="Image" src="https://github.com/user-attachments/assets/1c1e7024-57b6-4a1b-a402-fa589c8a0cd0" />

https://github.com/user-attachments/assets/f893b240-be7d-4c41-b819-35bb2b9ad990

### Setup - OS - Browser - Environment

any

### Priority

🔴 P0 : Critical

frol:
Resolved with #877

rknearn-qa:
@frol Partially reproducing

1. Choose e.g. USDC (Near Network)
2. Enter invalid address (e.g, bla-bla)
3. Observe UI state

Actual:
AVAX network is available
Wrong validation message displays

https://github.com/user-attachments/assets/38f0be1d-b350-4fe7-90ef-cfd0b03050f8

<img width="1365" height="681" alt="Image" src="https://github.com/user-attachments/assets/71edbc57-00f2-4f4e-97a2-5713058fcea7" />

* OS / Devices: Windows 11 (x64), Motorola G23 (Android 14), Iphone XR (iOs 18.7.6)
* Screen resolution(s): 1366 x 768, 1600 x 720, 1792×828
* Browser(s) + version(s): Chrome (149.0.7827.54), Mozilla Firefox (151.0.3), Brave (1.91.168), Safari (18.7.4), Edge (149.0.4022.52)
* Wallet: Meteor
* App environment: Staging ([trezu.app](https://testenv.trezu.app/staging-qa.sputnik-dao.near/payments))
* Near Network: Mainnet

frol:
@rknearn-qa So the issue now is that while it is not auto-selected, it is still available to be selected, and then it is a misleading error message, right?

rknearn-qa:
> So the issue now is that while it is not auto-selected, it is still available to be selected, and then it is a misleading error message, right?

@frol Sorry for delay, yeah, correct

We have opportunity to create request with invalid address and for some reason Destination Network will opt-in AVAX by default

rknearn-qa:
Retested
Fixed

[...]

PR #881, opened 2026-06-12
fix: correct recipient network autoselect

frol:
#880

- map "avax" bridge network name to EVM validation; unknown chains accepted any address and won autoselect (e.g. "b" -> Avalanche)
- wait for bridge networks to load before auto-picking, so near.com is not selected by default while real options are still loading
- never auto-pick a network whose address format cannot be validated
- reset network selection whenever the recipient address changes; auto-pick re-applies only on a single positive format match
- clear raw networkName form field when selection clears, so the submit gate cannot pass with an empty selector


Megha-Dev-19:
@frol I actually implemented these fixes in #877.
One difference: I don’t auto-select recipient network at all — even when only one option is available, user must select it manually so they know which network they selected.

frol:
@Megha-Dev-19 Ah, sorry, I missed that from the description of your PR. I will close mine as duplicate.

Issue #925, opened 2026-06-18
[BUG] Transfer details missing amount and destination network


Megha-Dev-19:
Some transfer entries/details do not show the amount and destination network, even though both should be visible.
https://trezu.app/meteor.sputnik-dao.near/requests/79

rknearn-qa:
Retested
Fixed

<img width="1362" height="680" alt="Image" src="https://github.com/user-attachments/assets/b1426e90-9794-4776-b4aa-80feadc46d9c" />

* OS / Devices: Windows 11 (x64), Motorola G23 (Android 14), Iphone XR (iOs 18.7.6)
* Screen resolution(s): 1366 x 768, 1600 x 720, 1792×828
* Browser(s) + version(s): Chrome (149.0.7827.54), Mozilla Firefox (151.0.3), Brave (1.91.168), Safari (18.7.4), Edge (149.0.4022.52)
* Wallet: Meteor
* App environment: Staging (trezu.app)
* Near Network: Mainnet

PR #927, opened 2026-06-18
fix: NEAR transfer token/network display

Megha-Dev-19:
#925 

Added a new `nearFt` parameter to the token metadata API.
NearBlocks does not return chain/origin metadata, so we now call NearBlocks only in safe cases:
- when chain metadata is not required, or
- when we already know the token is on NEAR (for example, Transfer proposals).

image

- Also fixed animation with onboarding progress bar

### DIFF
--- a/nt-be/src/handlers/balance_changes/history.rs
+++ b/nt-be/src/handlers/balance_changes/history.rs
@@ -1707,7 +1707,7 @@ pub async fn get_recent_activity(
 
     if !activity_token_ids.is_empty() {
         let chain_metadata_map =
-            fetch_tokens_with_fallback(&state, &activity_token_ids, true).await;
+            fetch_tokens_with_fallback(&state, &activity_token_ids, true, false).await;
         for change in &mut enriched_changes {
             if let Some(ref mut metadata) = change.token_metadata
                 && let Some(chain_meta) = chain_metadata_map.get(&change.token_id)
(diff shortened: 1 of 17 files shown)


Megha-Dev-19:
Here’s what’s going on with token metadata.

When we need to show a token in the UI (name, icon, network), we look it up from multiple sources in order until we find a match:

1. Our DB (`counterparties` table) — tokens we’ve seen before in balance changes. Has basics (symbol, icon, decimals), but no chain info.
2. Our hardcoded list (`tokens.json`) — curated Intents token list. Has chain info.
3. Chaindefuser API — live, broader version of the above. Also has chain info.
4. NearBlocks — last resort. Gives symbol/icon/decimals, but no chain data.

This works fine for truly native NEAR tokens. But many `.near` tokens are actually bridged. We had an issue with a SUI token in deposit earlier (I don’t remember the exact ID), which is why I changed this behavior #690. If NearBlocks resolves those tokens, the UI can incorrectly show `NEAR` as the network as it fallback to near network when chain data was missing.

So right now we skip NearBlocks whenever the UI needs chain info — it’s better to not show the token than show it wrong. That keeps bridged tokens correct, but brand-new native NEAR tokens (like Nexus) won't show up until they appear in one of the first three sources.

The hard part: for unknown tokens, we don’t have a reliable way to distinguish:
- native NEAR (safe to use NearBlocks)
- bridged token (NearBlocks would assign wrong chain)

That’s the part I’m trying to solve: how to handle this cleanly going forward.
@frol @ktarun1419

ktarun1419 (approved):

PR #947, opened 2026-06-20
Fix: Treat network fee as additive instead of deducting from payment amount (fixes #945)

### LINKED ISSUE #945: [BUG] Incorrect Payment Amount Display When Network Fee Is Applied
**User Story**
As a user, I want to make a test transaction using USDC on the Solana network from my treasury.

**Problem**
I have a treasury with a balance of 1.31794 USDC. When I try to create a payment for 0.1 USDC, I see an error saying that the amount is too low to cover the network fee.
However, we agreed that the network fee should be charged in addition to the amount entered by the user, not deducted from the payment amount itself. I expect the network fee to be taken from my available treasury balance, not from the payment amount I entered.

Expected Behavior
* Treasury balance: 1.31794 USDC
* Payment amount entered by user: 0.1 USDC
* Review screen shows:
    * Amount: 0.1 USDC
    * Network fee: 0.300092 USDC
* Total deducted from treasury: 0.400092 USDC

The transaction should be allowed because the treasury balance covers both the payment amount and the network fee.

**Corner Case**
If a user has 12 USDC in their treasury and attempts to withdraw the entire balance (12 USDC), an error should be displayed because there would be no remaining balance to cover the network fee.

In this case, the user should be required to either:
* Reduce the payment amount, or
* Leave enough balance to cover the network fee.

Image

Image

harshsingh07i:
**Description**
Resolves [#945](https://github.com/NEAR-DevHub/trezu/issues/945).

Previously, when creating a payment, the validation logic incorrectly checked if the payment amount itself was large enough to cover the network fee (e.g., trying to send 0.1 USDC with a 0.3 USDC fee would trigger an "amount too low" error). This PR fixes the validation logic so that the network fee is properly treated as an additional charge taken from the treasury balance, allowing users to send small payment amounts without artificial restrictions.

**Changes Made**
*   Updated `evaluateNetworkFeeCoverage` in `lib/intents-fee.ts` to correctly validate the total required amount (`amount + fee`) against the treasury balance instead of checking if the payment amount is larger than the fee.
*   Removed the `MAX_FEE_TO_RECIPIENT_RATIO` artificial limit in `hooks/use-intents-quote.ts` which improperly rejected valid single payments.
*   Added a new `insufficientBalanceForFee` label and error message in `messages/en.json` to handle the corner case where a treasury balance cannot cover both the payment amount and the network fee.


Megha-Dev-19:
Thanks for the PR @harshsingh07i! I built on your branch with a few tweaks:

- Kept additive network fees (EXACT_OUTPUT).
- Swapped blocking fee errors for non-blocking balance warnings (`amount + fee` vs balance).

Appreciate you picking this up!

Megha-Dev-19 (approved):

Issue #743, opened 2026-05-27
[Task] Add refresh and last updated to recent activity


ktarun1419:
### Background

Recent activity does not clearly show when it was last updated.

Users should be able to manually refresh the list and see the last updated time.

### User Story

As a user, I want to refresh recent activity and see when it was last updated, so that I know whether the data is current.

### Acceptance Criteria

- Add a refresh button to recent activity.
- Clicking refresh fetches the latest activity.
- Show last updated time near recent activity.
- Show loading state while refresh is running.
- Handle refresh errors without breaking the page.

rknearn-qa:
Implemented

<img width="1365" height="689" alt="Image" src="https://github.com/user-attachments/assets/f655cff4-bb2f-4b11-8002-ed5bf37ec29d" />
<img width="1365" height="678" alt="Image" src="https://github.com/user-attachments/assets/bcebbd10-e0cc-4862-ae4d-14717e9bfc49" />

* OS / Devices: Windows 11 (x64), Motorola G23 (Android 14), Iphone XR (iOs 18.7.6)
* Screen resolution(s): 1366 x 768, 1600 x 720, 1792×828
* Browser(s) + version(s): Chrome (149.0.7827.54), Mozilla Firefox (151.0.3), Brave (1.91.168), Safari (18.7.4), Edge (149.0.4022.52)
* Wallet: Meteor
* App environment: Staging (trezu.app)
* Near Network: Mainnet

Issue #573, opened 2026-04-30
Add partial data refresh button


olskik12:
We need to add a refresh button that allows users to update specific data on the page.
The refresh action should only update the following sections:
* Total balance
* Assets
* Recent transactions

Other parts of the page should remain unchanged and should not be reloaded.

[Figma](https://www.figma.com/design/UlUkb3wL7fQq0xNvmjFnEg/TREZU?node-id=7112-25550&t=7u14hYX4rxauvBBW-4)

https://github.com/user-attachments/assets/e182bfbe-7b30-4a37-9ff4-ffe53aab0e75

ori-near:
Hi @olskik12, can we please change "Update portfolio data" to "Update treasury data"?

olskik12:
@ori-near Yes, we can. 
@Megha-Dev-19 I have updated the UX Copy in the design

olskik12:
1. In the current implementation, the tooltip text on the button differs from the design — it needs to be updated.
2. We need to change the chart time range selector; it should be a dropdown menu selector.
3. Also, for the token selector, we need to update the style: instead of a border, it should have a background like the buttons.

All of this should be implemented for both types of treasury

[Figma](https://www.figma.com/design/UlUkb3wL7fQq0xNvmjFnEg/TREZU?node-id=7112-25550&t=JRnHEWzy7viRyyS7-4)

<img width="950" height="510" alt="Image" src="https://github.com/user-attachments/assets/22e10abc-ea1b-40af-aaa5-7afb78eca6a5" />

<img width="889" height="644" alt="Image" src="https://github.com/user-attachments/assets/6d5de656-7e6d-4934-8a5e-f4cf29c7c0fd" />

olskik12:
Hi @Megha-Dev-19 @ktarun1419 I see that you moved this to QA In Progress, but I don’t see the UI improvements I mentioned.
Also, right now we don’t have a loading state when data is being fetched (I showed this in the video, and we also have a skeleton for it)

rknearn-qa:
1. Missing UI skeleton with periods - #982
2. UI not triggers after Refresh - #981
3. Small UI issue (may skip) - #983

* OS / Devices: Windows 11 (x64), Motorola G23 (Android 14), Iphone XR (iOs 18.7.6)
* Screen resolution(s): 1366 x 768, 1600 x 720, 1792×828
[...]

PR #949, opened 2026-06-20
feat: refresh history button

ktarun1419:
#951 #743


nt-fe/hooks/use-treasury-queries.ts
```diff
@@ -355,6 +362,65 @@ export function useRecentActivity(
     });
 }
 
+export function useConfidentialHistoryRefreshStatus(
+    accountId: string | null | undefined,
```
Megha-Dev-19, line 366: why is the account id null? don't we only call it on treasury page, so it will always be there?

nt-fe/messages/de.json
```diff
@@ -373,6 +373,16 @@
...
+        },
+        "refresh": {
+            "ariaLabel": "Refresh confidential history",
+            "refreshing": "Refreshing…",
+            "lastUpdated": "Last updated {time}",
+            "notUpdated": "Not updated yet",
+            "justNow": "now",
+            "moments": "moments",
+            "checking": "Checking refresh status…",
+            "unavailable": "Refresh status unavailable"
```
Megha-Dev-19: in all translations it's in english, please update it.

Megha-Dev-19:
@ktarun1419 we have this #573 ticket, can we add this refresh button for the whole page? If not, please confirm with olha once about your implementation. Thanks!

nt-fe/hooks/use-treasury-queries.ts
ktarun1419: In practice this is the treasury account id and should usually be present on this page, but the hook accepts null/undefined to match `useTreasury().treasuryId` and our existing React Query pattern. 

We call hooks unconditionally and let `enabled: !!accountId` disable the query until the route param exists. Several nearby hooks use the same `string | null | undefined` shape.

ktarun1419:
> @ktarun1419 we have this #573 ticket, can we add this refresh button for the whole page? If not, please confirm with olha once about your implementation. Thanks!

In that ticket, @olskik12  mentioned a partial data refresh for a specific section, as the balance refresh in the referenced ticket. We are following the same approach for transaction history.

Adding a single refresh button for the whole page may not be feasible because the page depends on data from multiple sources. For recent history specifically, we need to run the full flow, and there is also a cooldown period involved.

olskik12:
@ktarun1419 @Megha-Dev-19 Could you please provide a screenshot of how this is implemented in the UI? I’m having trouble understanding how it’s supposed to look just from the code

The user flow should match the design: when the user clicks the Refresh button, only the balance, assets, and recent transactions should be refreshed. Pending requests do not need to be refreshed

ktarun1419:
Screenshot 2026-06-22 at 4 49 37 PM

currently this is the design for refresh of recent transaction only

olskik12:
Thanks for the screenshot! We don’t need a separate refresh button in a different place. Please add the refresh button as shown in the design, in the top “Total Balance” section. It should refresh the Total Balance, Assets, and Recent Transactions. Also, the tooltip text should match the design.

Megha-Dev-19 (approved):

PR #975, opened 2026-06-25
feat: analytic dashboard

ktarun1419:
#976 

- added usd value at the time of confidential snapshot
- added analytics table per treasury per month required by Ori for better data visualisation

### DIFF
--- a/nt-be/src/handlers/intents/confidential/gold/snapshots/worker.rs
+++ b/nt-be/src/handlers/intents/confidential/gold/snapshots/worker.rs
@@ -51,6 +51,22 @@ pub async fn snapshot_confidential_dao_balances(state: &AppState, dao_id: &str) 
 
     let defuse_map = get_defuse_tokens_map();
     let snapshot_at = Utc::now();
+    let live_balances: Vec<(String, String)> = live_balances.into_iter().collect();
+    let live_assets: Vec<String> = live_balances
+        .iter()
+        .map(|(asset, _)| asset.clone())
+        .collect();
+    let latest_prices = match state
+        .price_service
+        .get_cached_tokens_latest_price(&live_assets)
+        .await
+    {
+        Ok(prices) => prices,
+        Err(e) => {
+            tracing::warn!("{} batch snapshot price lookup failed: {}", dao_id, e);
+            std::collections::HashMap::new()
+        }
+    };
     let mut rows: Vec<SnapshotRow> = Vec::with_capacity(live_balances.len());
     let mut seen_assets = std::collections::HashSet::with_capacity(live_balances.len());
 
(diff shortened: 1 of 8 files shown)


nt-be/src/handlers/intents/confidential/gold/snapshots/worker.rs
```diff
@@ -77,12 +77,34 @@ pub async fn snapshot_confidential_dao_balances(state: &AppState, dao_id: &str)
             acc * BigDecimal::from(10u32)
         });
         let balance = &raw_balance / &scale;
+        let price_usd = match state.price_service.get_latest_cached_price(&asset).await {
+            Ok(Some(price)) => {
```
Megha-Dev-19: can we use `get_cached_tokens_latest_price` as it does batch lookup instead of just one asset.

Megha-Dev-19 (commented):
What will we do for the existing data? do we want to add a backfill for it?
And I remember when Peter added the fetch for DeFiLlama prices we hit the limit on staging, what's the current limit we have?

ktarun1419:
> What will we do for the existing data? do we want to add a backfill for it? And I remember when Peter added the fetch for DeFiLlama prices we hit the limit on staging, what's the current limit we have?

Not adding a historical backfill in this PR. I’ll handle that after the public pricing refactor.

On rate limiting: DeFiLlama does not clearly document its rate limits. I tested local requests per second, per minute, and per hour, and they worked fine.

This PR does not change the price sync interval. It was 60 seconds before and remains 60 seconds here.

nt-be/src/handlers/intents/confidential/gold/snapshots/worker.rs
ktarun1419: Good catch, I extracted the `assets`, did a batch lookup, and used that.

Fixed

Megha-Dev-19:
Yeah, earlier it was 60 seconds, but it only ran when price for that day didn’t exist. Now we always run it after 1 minute.
One thing I’m thinking about: right now we update the same DB row and store only one price per day per asset.

When we add public support later, we’ll likely need the same historical price data per asset/day. Should we start storing it properly in DB now?

ktarun1419:
> Yeah, earlier it was 60 seconds, but it only ran when the price for that day didn’t exist. Now we always run it after 1 minute. One thing I’m thinking about: right now, we update the same DB row and store only one price per day per asset.
> 
> When we add public support later, we’ll likely need the same historical price data per asset/day. Should we start storing it properly in DB now?

Yes, that’s fair. As I mentioned before, I tested the current DeFiLlama calls locally with all mapped tokens: 60 sequential requests, each including the full mapped token set (Worked fine). Since we only fetch distinct mapped DeFiLlama IDs, the request size is bounded by our token mapping, not by the number of DAOs or snapshots. I also tested per-second, per-minute, and per-hour calls, and they worked fine.

For the revamp, I agree. We should store pricing more formally as part of the public/token flow refactor. My thinking is to introduce a dedicated tokens/assets table and reference it by FK wherever we store balances, events, and prices. Prices can then be keyed by the same asset FK plus date/time.

For this PR, I kept the scope small to unblock Ori’s analytics requirement for confidential AUM.

Megha-Dev-19:
let's keep a check on staging once we merge it, to see how it goes @ktarun1419

Megha-Dev-19 (approved):

Issue #980, opened 2026-06-25
[BUG] Confidential payment/exchange shows Executed too early (should reflect Intents status)


Megha-Dev-19:
Right now, once a confidential payment/exchange request is approved, we show it as Executed immediately.
But on Intents side, it can still be Processing or Failed.
We should update the UI to show the correct status (Processing / Executed / Failed), same as we do for public treasury requests.

rknearn-qa:
Blocked by #1031

rknearn-qa:
Retested
Fixed: implemented

<img width="1365" height="695" alt="Image" src="https://github.com/user-attachments/assets/5b8d6c49-ef45-42c1-b372-db36db0e9619" />

* OS / Devices: Windows 11 (x64), Motorola G23 (Android 14), Iphone XR (iOs 18.7.6)
* Screen resolution(s): 1366 x 768, 1600 x 720, 1792×828
* Browser(s) + version(s): Chrome (149.0.7827.54), Mozilla Firefox (151.0.3), Brave (1.91.168), Safari (18.7.4), Edge (149.0.4022.52)
* Wallet: Meteor
* App environment: Staging (trezu.app)
* Near Network: Mainnet

PR #1004, opened 2026-06-30
Feat: confidential swap status

Megha-Dev-19:
#980 
- Added confidential swap status using `bronze_confidential_history_events`
- Updated confidential link to use `mask` explorer link instead of nearblocks
- Fixed confidential loading UI for expanded request details page


nt-be/src/handlers/intents/swap_status.rs
```diff
@@ -138,26 +143,100 @@ pub fn extract_quote_data(full_response: &FullSwapStatusResponse) -> Option<Quot
...
     let http_client = state.http_client.clone();
     let oneclick_jwt_token = state.env_vars.oneclick_jwt_token.clone();
     let oneclick_api_url = state.env_vars.oneclick_api_url.clone();
+    let db_pool = state.db_pool.clone();
 
     let result = state
         .cache
         .cached(CacheTier::ShortTerm, cache_key, async move {
+            if let Some(dao_id) = dao_id.as_deref() {
+                let is_confidential = is_confidential_dao(&db_pool, dao_id).await.map_err(|e| {
```
frol: Overall, I don't like this ad-hoc mix of confidential and public data and lack of structure. Even the naming becomes inconsistent/unaligned: e.g. `fetch_confidential_swap_status` & `fetch_swap_status_response` (no "public", and extra "_response"), and the if statement does early return instead of a clear `if {} else {}` that better communicates that the paths are of the same level and not that one is just a shortcut of the full path.

nt-be/src/handlers/intents/swap_status.rs
```diff
@@ -138,26 +143,100 @@ pub fn extract_quote_data(full_response: &FullSwapStatusResponse) -> Option<Quot
...
+/// Read confidential swap status from bronze. `None` means not yet ingested.
+pub async fn fetch_confidential_swap_status(
+    pool: &sqlx::PgPool,
+    dao_id: &str,
+    deposit_address: &str,
+) -> Result<Option<SimplifiedSwapStatusResponse>, (StatusCode, String)> {
+    let row = sqlx::query_as::<_, (String, chrono::DateTime<chrono::Utc>)>(
+        r#"
+        SELECT status, created_at_external
+        FROM bronze_confidential_history_events
```
frol, line 171: @ktarun1419 I am not completely sure I like the idea of using bronze data, though since it is only using it to report the status and does not leak the actual data columns, I think it is fine. What do you think?

ktarun1419: I think a better approach would be to copy `status` into `gold_confidential_history_events` during projection and have the API read from gold.

Reading from bronze here probably won’t break the medallion flow since it’s read-only, but it does create a side path around the read model. If status is something the frontend needs, gold should expose it. It may also be useful in the future.

Megha-Dev-19: The catch is that gold is a SUCCESS-only projection — `load_bronze_suffix` filters `he.status = 'SUCCESS'`, and `delete_stale_gold_rows` drops any gold row whose bronze event stops being `SUCCESS`.

This endpoint's whole purpose is to report the *non-success* lifecycle states — `PROCESSING` / `PENDING_DEPOSIT` while in flight, and terminal `REFUNDED` / `FAILED`. If we read status from gold:

- processing/pending → no gold row → we default to `PROCESSING` ✅
- **refunded/failed → never enters gold → would show `PROCESSING` forever** ❌

So copying `status` into gold doesn't actually work unless we also start projecting non-SUCCESS rows into gold, which would mean gold is no longer a clean SUCCESS-only read model and we'd have to add `status`/`transaction_type` guards to every activity-feed and balance query that currently assumes SUCCESS-only. That felt like a much larger change than this endpoint warrants.

Given it's read-only, I'd propose keeping the bronze read for now. If we later want status in the read model.

ktarun1419: Yes we need to do that. For pending rows, the rule should be:

If the bronze row is still `PENDING` and the linked intent/proposal has already been submitted, we can create/update a gold row with `status = pending` so the UI can show that the request is in progress.

But this row should be status-only:
- no balance changes
- no ledger mutation
- no finalized amount history

Once bronze becomes `SUCCESS`, the gold projector will automatically update the same row into a normal finalized history row and apply the balance changes. If it becomes `FAILED` / `REFUNDED`, we only update the status and still avoid balance changes.

Yes we can use this for now. I already implemented this in the public flow and will update this later on

ktarun1419 (approved):

Issue #1005, opened 2026-06-30
[TASK] Show deposit sender and transaction link in confidential treasury activity


Megha-Dev-19:
When a deposit comes into a confidential treasury via NEAR Intents, the 1Click API now returns the actual on-chain sender address and transaction hash in quoteTransactions.
The activity feed should show the sender and the transaction link.

rknearn-qa:
Implemented

<img width="1365" height="695" alt="Image" src="https://github.com/user-attachments/assets/65b70f96-6afc-4c6a-a329-ad2794ec49f3" />
<img width="1358" height="684" alt="Image" src="https://github.com/user-attachments/assets/b709e71b-e798-4ba3-b7d5-f740a3a72834" />
<img width="1317" height="689" alt="Image" src="https://github.com/user-attachments/assets/a4a50a97-0fc6-4b87-9f1f-5768bbb0496b" />

* OS / Devices: Windows 11 (x64), Motorola G23 (Android 14), Iphone XR (iOs 18.7.6)
* Screen resolution(s): 1366 x 768, 1600 x 720, 1792×828
* Browser(s) + version(s): Chrome (149.0.7827.54), Mozilla Firefox (151.0.3), Brave (1.91.168), Safari (18.7.4), Edge (149.0.4022.52)
* Wallet: Meteor
* App environment: Staging (trezu.app)
* Near Network: Mainnet

Issue #966, opened 2026-06-23
[UX] Show recipient transaction link for payments


frol:
### Background

When we make a payment to a recipient on the external blockchain, we need to be able to provide the link to the transaction where the recipient receives the transfer.

### User Story

As a treasury manager, I need to be able to get the transfer link from the executed request or from the recent activity and put it into my accounting system or provide it to the recipient as a proof of payment.

### Acceptance Criteria

* Proposal details has the link to the external (recipient) block explorer that has the transfer amount deposited to the recipient’s account
* Recent transaction modal has the link to the external (recipient) block explorer that has the transfer amount deposited to the recipient’s account
* There should be individual links for the bulk payment transfers

olskik12:
Hi @Megha-Dev-19 @ktarun1419 I have a design for the case when a user makes a payment on an external blockchain. 

**Origin Chain** - NEAR Intents link
**Destination Chain** - network transaction destination that received this transaction
 
Individual Payment - [Figma](https://www.figma.com/design/UlUkb3wL7fQq0xNvmjFnEg/TREZU?node-id=9110-109476&t=728FSA4PNg4n7Nrk-4)

<img width="844" height="602" alt="Image" src="https://github.com/user-attachments/assets/61e408b8-c8dd-430e-a8c3-c807f55049a9" />

Bulk Payment - [Figma](https://www.figma.com/design/UlUkb3wL7fQq0xNvmjFnEg/TREZU?node-id=9202-81708&t=728FSA4PNg4n7Nrk-4)

<img width="801" height="573" alt="Image" src="https://github.com/user-attachments/assets/7ac6bb49-50d8-4841-abec-5c7e9596d7ae" />

Recent transaction (only for "Sent" details)

<img width="629" height="448" alt="Image" src="https://github.com/user-attachments/assets/7a1f86c3-ba2d-49c1-a68a-b55d9e931e75" />

PR #1020, opened 2026-07-01
feat: surface origin tx + sender for confidential deposits

Megha-Dev-19:
#1005 
- Save the sender and transaction hash from 1Click's `quoteTransactions` onto confidential gold events (`sender_address`, `deposit_tx_hash`) using a new migration.
- For confidential **deposits**, show who sent the funds and a working link to the on-chain transaction (labeled "Deposit Transaction").
image

- We only do this for deposits. **Sends** happen privately inside intents, so their hash isn't a real on-chain transaction, and **exchanges** don't include this data, from what I have gathered in our testing treasuries, Tarun will ask more about it from the team, as doc doesn't provide any data about it.
- The From/To columns in the activity table now show a short address, with the same hover card as the `User` component.
- Deleted warnings doc as we want to make it internal.


nt-be/src/handlers/intents/confidential/gold/history_events/classify.rs
```diff
@@ -392,6 +421,8 @@ pub(crate) fn project_row(
         proposal_execution_transaction_hash: row.proposal_execution_transaction_hash.clone(),
         quote_created_at: row.created_at_external,
         proposal_created_at: row.proposal_created_at,
+        sender_address,
+        deposit_tx_hash,
```
frol: I am getting more and more confused about `recipient`, `counterparty`, `deposit_address`, `sender_address`. Since the sender or receiver is always the `dao_id`, there should be a single field that has the sender (in case of a deposit) or the receiver (in case of the transfer).

@ktarun1419 @Megha-Dev-19 I have hard time understanding why we need 4 fields. Please, clarify it with comments/better names or (preferably) refactor.

frol (changes requested):

nt-be/src/handlers/intents/confidential/gold/history_events/classify.rs
ktarun1419: Agreed, we should simplify this.

Field meanings/usage:
- `dao_id`: our DAO/treasury account. This is implicit in the page/API context.
- `recipient`: raw 1Click recipient. Mainly useful for classification: sent vs deposit vs exchange.
- `counterparty`: the field exposed to the frontend as the other side of the activity. This is what we use for From/To display, so it should hold the sender for deposits and the receiver for sends.
- `deposit_address`: temporary 1Click deposit address used to fund the quote. Useful for mapping the data with confidential_intents, but not a From/To account.
- `sender_address`: not needed as a separate field if we store the deposit sender in `counterparty`.

So I think `counterparty` should be the single frontend-facing field, and `sender_address` can be removed.

@frol @Megha-Dev-19

frol:
@Megha-Dev-19 @ktarun1419 If we would take into account #966, which would require us to also store the destination chain transaction hash, would it change how we implement this PR? Should we consider something like "external_tx_hash" which would be a deposit tx hash for deposits from public chains and it would be the recipient tx hash for the payments? Though we should also take into account bulk payments, which would have a list of recipient tx hashes.

I approve this PR as is as it resolves the original issue, but if the cost of future refactoring would be higher (would require a migration of some sort), then I would recommend to address it before we merge this PR.

frol (approved):

ktarun1419:
> @Megha-Dev-19 @ktarun1419 If we would take into account #966, which would require us to also store the destination chain transaction hash, would it change how we implement this PR? Should we consider something like "external_tx_hash" which would be a deposit tx hash for deposits from public chains and it would be the recipient tx hash for the payments? Though we should also take into account bulk payments, which would have a list of recipient tx hashes.
> 
> I approve this PR as is as it resolves the original issue, but if the cost of future refactoring would be higher (would require a migration of some sort), then I would recommend to address it before we merge this PR.

@frol  I’d keep `deposit_tx_hash` as-is and handle #966 separately.

`deposit_tx_hash` is specifically the source/deposit tx and is always singular. #966 is about recipient-side tx links for payments, and bulk payments need multiple recipient tx hashes, so a single `external_tx_hash` would not model that well anyway.
also, 1Click does not expose payment hashes yet, so we will later design the recipient-side schema once we know the data shape.

also, this table is a gold projection. Bronze keeps the raw payload, so if we need to reshape this later it should be an additive migration + re-projection only.

Issue #989, opened 2026-06-26
[BUG] Trezu - Confidential Treasury: NEAR-Only tokens return 401 Authentication Error for Non-Members on Deposit Page


rknearn-qa:
### Summary

On Confidential Treasuries, users who are not logged in or are not treasury members cannot select deposit options for tokens that are available only on the **NEAR network**.

Selecting one of these tokens results in a **401 Unauthorized** error instead of displaying a deposit address or an appropriate permission message.

The issue affects more than **30 NEAR-only tokens**, making a significant portion of available deposit assets unusable for external contributors. Relates to #988 #778 

### Preconditions:

* Confidential Treasury exists (https://testenv.trezu.app/confidential-staging.sputnik-dao.near)
* User is **not logged in** or is **not a treasury member**

### Steps To Reproduce

1. Open the **Deposit** page of a Confidential Treasury
2. Select any token that is available **only on the NEAR network**
3. Observe the UI and network requests

### Actual Result:

UI displays:

```text
Request failed with status code 401
```

Network response:

```text
Authentication required for confidential treasury
```

As a result:

* Deposit cannot be completed
* Deposit address is not generated
* The issue reproduces across 30+ NEAR-only tokens

<img width="1365" height="681" alt="Image" src="https://github.com/user-attachments/assets/3aaf7fe2-83ce-4198-8289-725019fe20c9" />
<img width="1361" height="679" alt="Image" src="https://github.com/user-attachments/assets/16441984-0d4b-4775-8017-3cc4ebc8f9c2" />
<img width="866" height="690" alt="Image" src="https://github.com/user-attachments/assets/7adef7ed-06de-4f41-aa58-8a4db633f986" />
<img width="906" height="682" alt="Image" src="https://github.com/user-attachments/assets/614ff20a-c6c6-4f5c-9e4a-15def38f571a" />
<img width="915" height="684" alt="Image" src="https://github.com/user-attachments/assets/99b14dc6-83bc-4e34-8add-82678a7dc447" />
<img width="950" height="687" alt="Image" src="https://github.com/user-attachments/assets/fc4bc71a-702a-4aca-9f0c-8320ddae8ed3" />
<img width="933" height="685" alt="Image" src="https://github.com/user-attachments/assets/37fbb704-6e2a-4ced-8ae4-5a1ea6048153" />
[...]

PR #1022, opened 2026-07-02
Fix: don't auto-select network for confidential guest users in deposit modal

Megha-Dev-19:
#989 
In the deposit modal, when an asset has only one network, we auto-select that network. But for guests / non-logged-in users on confidential treasuries, network selection is disabled.


frol (approved):
It is a good quick fix. In the future, we may consider allowing the backend to generate deposit addresses for non-authenticated / non-member users.

PR #1023, opened 2026-07-02
fix: make creation idempotent, resumable, and auto-retrying

Megha-Dev-19:
#1024

Treasury creation used to get stuck when an RPC call timed out — the DAO would be created but ownership/confidential setup never finished, leaving the user's chosen name "reserved" and unusable.

- **Idempotent & resumable creation:** each step now checks on-chain/DB state before acting, so a re-run skips work that's already done (DAO create, add public key, 1Click auth) instead of duplicating it or failing.
- **Automatic retries:** the creation flow is wrapped in a bounded auto-retry loop (4 attempts, exponential backoff) that resumes from the last completed step on transient RPC/transport errors. Runs in the background, so it keeps going even if the client disconnects.
- **Concurrency guard:** a per-account Postgres advisory lock prevents two create calls for the same treasury from running at the same time.
- **More RPC endpoints + retries:** added fallback mainnet RPC endpoints and raised per-endpoint retries (7) for better resilience.

### DIFF
--- a/nt-be/src/app_state.rs
+++ b/nt-be/src/app_state.rs
@@ -16,6 +16,48 @@ use crate::{
     },
 };
 
+/// Per-endpoint RPC retries before near-api fails over to the next endpoint
+/// (near-api's default is 5). Bumped so transient RPC hiccups are absorbed
+/// before giving up on an endpoint.
+///
+/// NOTE: near-api only fails over on retryable errors (HTTP 408/429/5xx from a
+/// reachable server); hard connection errors and client timeouts are treated as
+/// `Critical` and abort immediately without failover. Timeout/connection
+/// resilience therefore comes from the multiple endpoints.
+const RPC_ENDPOINT_RETRIES: u8 = 7;
+
+/// Build the default mainnet RPC endpoints with provider-level failover.
+///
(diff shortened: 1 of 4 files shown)


frol (changes requested):
I fear that (even with the old implementation) if the user hits the creation endpoint twice (e.g. there is a network glitch and the requests is sent twice from the frontend), then the process could get on its own toes. We need some sort of a lock/status in the database (which is our shared state).

Megha-Dev-19:
@frol The double submit race is handled here in two layers.

First, there's a per-account lock in our shared DB state. Right at the start of `run_creation`, before any on-chain work, we grab a Postgres advisory lock keyed by the account id. If a second request comes in for the same treasury (a double-send from a network glitch), it can't get the lock and gets rejected right away instead of running alongside the first one:

I went with an advisory lock instead of a `status` column as it auto-releases if the process crashes, so we don't end up stuck in a permanent "in_progress" state that needs manual cleanup.

Second, even if two calls somehow interleaved, `run_creation_inner` starts with `classify_treasury_account`, which checks the on-chain state and skips the create (and confidential steps) if the DAO already exists. So a duplicate can't make a second DAO or double-charge.

frol (approved):

Issue #994, opened 2026-06-29
[Task] Integrate Apalis worker queue and SSE event bus infrastructure


ktarun1419:
### Background

Add a generic background job processing layer using Apalis and a server-sent events flow for pushing async backend updates to the frontend.

This should be implemented as reusable infrastructure and should not be tied to any specific feature. Future flows should be able to enqueue jobs, process them in workers, publish completion/failure events, and notify the frontend without requiring manual page refreshes.

### User Story

As a user, I need to see my transaction history instantly, without manually refreshing the page.

### Acceptance Criteria

1. Add Apalis with Postgres-backed job storage.
2. Create a generic worker setup that can run independently from the API server.
3. Add support for enqueueing typed background jobs from API/service code.
4. Add retry, backoff, and failure handling for jobs.
5. Add basic job status tracking.
6. Add a shared backend event bus for async updates.
7. Add an SSE endpoint for frontend clients to subscribe to backend events.
8. Publish SSE events when relevant async jobs are completed or failed.
9. Ensure frontend can listen to events and trigger refetch/invalidation logic.

rknearn-qa:
Blocked by https://github.com/NEAR-DevHub/trezu/issues/1127

rknearn-qa:
Implemented
Move to Done

Issue #818, opened 2026-06-05
[Task] Refactor transactions history for Public treasuries


frol:
### Background

#641 paved the road for the reliable data pipeline with data traveling through the Bronze -> Silver -> Gold stages.

We need to now refactor the data sources for the public treasuries as well.

### User Story

• As a user, I want my public treasury transaction history to be 100% accurate, with exact execution timestamps and correct quote-time USD values that would allow advanced analytics, filtering, searching.
• As an operator, I want a resilient data pipeline that safely stores raw events so we can easily backfill data during downtime without losing history.

### Acceptance Criteria

* Gold historical records must have all the necessary information for the UI to display it without a need for additional ad-hoc queries or transformations
* The delay between the moment of event being initiated to it getting to our Gold stage should be as short as possible (<1 minute, but ideally <5 seconds, and for user-initiated actions, ideally have optimistic previews available immediately).

ori-near:
Currently blocked on external dependency from Allium (missing data in their API).

ktarun1419:
one thing that we need to take into account on the public treasuries is the native NEAR balance of the account which grows not only with user deposits, but also due to:

1. Gas rewards for contract function calls usage (30% of the gas fees associated with the function call execution are rewarded to the smart contract account, so every add_proposal / act_proposal gives something like 0.0001 NEAR - useless amount, but could be confusing to the users)

2. Storage compression for the proposals which we transfer as a separate transaction before relaying add_proposal

We deduct (2) from the total balance and on the charts and don’t show in the recent activity. We don’t handle (1) yet at all

Overall, the idea was that we didn’t want to show non-zero balance right after treasury creation as that may create confusion and unnecessary interest to drain Trezu for those micro amounts

rknearn-qa:
Blocked by #1127

rknearn-qa:
Implemented
Move to Done

PR #1028, opened 2026-07-02
Revamp/public txn history

ktarun1419:
#994 #818 

This PR adds:

- Adds a Medallion-style public history pipeline: Bronze ingestion, Silver transfer-leg normalization, and Gold frontend-ready history projection.
- Uses Goldsky as the activity trigger to enqueue targeted RefreshLatest jobs, reducing reliance on polling and making new treasury activity visible faster.
- Adds proposal receipt linking and a periodic reconciler to safely converge proposal statuses missed during receipt processing.
- Adds an in-process SSE event bus and /events endpoint so the frontend can refetch treasury data when projections change instead of polling (can be use at other places to avoid polling and get fresh data).
- Adds token price ingestion with canonical token registry, 5-minute USD price series, and monthly Postgres range partitions.
- Adds shared Postgres-backed Apalis job queue infrastructure with deduped in-flight jobs and dedicated public history workers for RefreshLatest and BackfillPage.


frol (approved):
It looks good and I think we should merge it now and keep it testing on staging

Issue #1024, opened 2026-07-02
[BUG] Treasury creation gets stuck on RPC failures, reserving the DAO name


Megha-Dev-19:
## Problem

When creating a treasury, an RPC timeout or transient network error can halt the process partway through. This is especially common in the confidential flow: the DAO gets created on-chain, but ownership transfer / confidential
setup never completes.

The result is a broken, half-created treasury:
- The user's chosen name is now "reserved" (the account exists), so they can't reuse it.
- The treasury isn't usable because setup didn't finish.
- There's no automatic way to recover — the user is stuck.

rknearn-qa:
Retested
@Megha-Dev-19 Still reproducing with next conditions partially:

Steps:

1. Open https://testenv.trezu.app/create
2. Open DevTools → Console
Paste this before clicking Create:
```
(() => {
  if (window.__bug1024Sim) {
    console.log("Bug #1024 simulation already active");
    return;
  }
  window.__bug1024Sim = true;
  const realFetch = window.fetch.bind(window);
  window.fetch = async (input, init) => {
    const url = String(input);
    if (url.includes("/api/treasury/create-stream")) {
      const body = init?.body ? JSON.parse(init.body) : {};
      console.log("[SIM] create-stream for", body.accountId);
      const events = [
        `data: {"step":"creating_dao","status":"in_progress"}\n\n`,
        `data: {"step":"creating_dao","status":"completed"}\n\n`,
        `data: {"step":"adding_public_key","status":"in_progress"}\n\n`,
        `data: {"step":"error","status":"error","message":"Simulated RPC timeout after DAO creation"}\n\n`,
      ];
      const stream = new ReadableStream({
        start(controller) {
          for (const chunk of events) {
            controller.enqueue(new TextEncoder().encode(chunk));
          }
          controller.close();
        },
      });
      return new Response(stream, {
        status: 200,
        headers: { "Content-Type": "text/event-stream" },
      });
    }
    if (url.includes("/api/treasury/check-handle-unused")) {
      console.log("[SIM] name is reserved");
      return new Response(JSON.stringify({ unused: false }), {
        status: 200,
        headers: { "Content-Type": "application/json" },
      });
    }
    return realFetch(input, init);
  };
  console.log("Bug #1024 simulation enabled");
})();
```

3. Fill the form:

Confidential
Any treasury name
Account name

4. Connect wallet → Create Treasury
5. Reload page
6. Use the same treasury account name
[...]

PR #1033, opened 2026-07-03
feat: Treasury creation sweeper

Megha-Dev-19:
#1024
The previous PR made treasury creation idempotent, resumable, concurrency-safe, and retried transient errors in-request. But if all in-request retries failed (or the process died mid-flow), the treasury was still left half-created until someone intervened.

This PR adds durable, automatic recovery so those half-created treasuries always get finished — and the user isn't stranded in the meantime.

**Persist creation intent**
- New `treasury_creation_requests` table stores the full request. Status lifecycle: `in_progress` while an attempt runs → `pending` if it fails part-way → row **deleted** on success → `failed` only after max attempts. For confidential DAOs this is the only way to recover the intended member policy, since it can't be read back from chain.

**Background sweeper**
- A worker resumes any `pending` creation by re-running the (already idempotent) flow. It's **woken immediately** when an attempt fails, so recovery starts within moments, and also **polls** as a fallback for crashed/abandoned attempts (reclaims stale `in_progress` rows). Gives up after max attempts → marks `failed` + Telegram alert.

**In-session frontend recovery**
- After a resumable error, the progress modal stays open, holds a spinner on the stalled step, and re-drives the idempotent flow until it completes — then redirects the user straight into their treasury. Permanent errors (name taken / creation disabled) go straight to the waitlist.

### DIFF
--- a/nt-be/migrations/20260703000001_create_treasury_creation_requests.sql
+++ b/nt-be/migrations/20260703000001_create_treasury_creation_requests.sql
@@ -1,0 +1,30 @@
+-- Persist treasury creation intent so a background sweeper can resume/finish
+-- any creation that failed part-way (e.g. RPC timeout after the DAO was created
+-- but before confidential ownership handoff). The intended policy (members +
+-- thresholds) is NOT recoverable from chain for a half-created confidential
+-- DAO, so we store the original request here.
+CREATE TABLE treasury_creation_requests (
+    account_id           TEXT PRIMARY KEY,
+    name                 TEXT NOT NULL,
+    payment_threshold    SMALLINT NOT NULL,
+    governance_threshold SMALLINT NOT NULL,
+    governors            TEXT[] NOT NULL,
+    financiers           TEXT[] NOT NULL,
+    requestors           TEXT[] NOT NULL,
+    is_confidential      BOOLEAN NOT NULL DEFAULT false,
+    -- 'in_progress' → an attempt is actively running; sweeper leaves it alone
+    --                 until it goes stale (crash/restart safety net).
+    -- 'pending'     → attempt failed part-way; eligible for sweeping right away.
+    -- 'failed'      → gave up after too many attempts (terminal, needs a human).
+    -- Rows are DELETED on success, so the table only holds unfinished creations.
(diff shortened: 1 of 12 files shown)


nt-be/migrations/20260703000001_create_treasury_creation_requests.sql
```diff
@@ -0,0 +1,30 @@
...
+    governors            TEXT[] NOT NULL,
+    financiers           TEXT[] NOT NULL,
+    requestors           TEXT[] NOT NULL,
+    is_confidential      BOOLEAN NOT NULL DEFAULT false,
+    -- 'in_progress' → an attempt is actively running; sweeper leaves it alone
+    --                 until it goes stale (crash/restart safety net).
+    -- 'pending'     → attempt failed part-way; eligible for sweeping right away.
+    -- 'failed'      → gave up after too many attempts (terminal, needs a human).
+    -- Rows are DELETED on success, so the table only holds unfinished creations.
+    status               TEXT NOT NULL DEFAULT 'in_progress',
```
ktarun1419: we can use enums for status (not a blocker)

nt-be/src/handlers/treasury/create.rs
```diff
@@ -482,6 +491,31 @@ async fn run_creation(
...
+            {
+                tracing::warn!("Failed to delete creation request for {}: {}", treasury, e);
+            }
+        }
+        Err(evt) => {
+            // Flip to `pending` so the sweeper picks it up, then wake the
+            // sweeper right away instead of waiting for its next poll tick.
+            let message = evt.message.as_deref().unwrap_or_default();
+            if let Err(e) =
+                creation_requests::mark_creation_pending(&state.db_pool, treasury.as_str(), message)
```
ktarun1419: We should handle ExistingDaoState::Taken as a terminal case and avoid retrying.

ktarun1419 (approved):

nt-be/src/handlers/treasury/create.rs
Megha-Dev-19: yeah good catch, will update

PR #1045, opened 2026-07-06
Fix zero network fee on confidential bulk payments to other chains

frol:
Paying a confidential bulk to Solana (or any non-NEAR chain) showed a zero network fee on the review screen, skipped the per-recipient fee padding, and recipients received less than entered - the withdrawal fee was deducted from their leg.

Two compounding causes, both silent:
1. The fee-estimation token override only replaced `address` with the destination asset id and kept the SOURCE token's `network`. For a near-network source token, isIntentsCrossChainToken() then classified the transfer as not cross-chain and returned a null fee without ever consulting the estimator.
2. estimateIntentsNetworkFee() returned a zero fee when the representative recipient address failed validation against the (wrongly sourced) destination chain - pricing an impossible withdrawal as free.

Fixes:
- upload step passes the destination network (and drops the source deployment's minWithdrawalAmount) in the fee-estimation token
- estimateIntentsNetworkFee throws on chain/address mismatch instead of returning zero
- fee-estimation failures now block the confidential cross-chain review step (previously only public bulk surfaced them)
- last-line guard in onSubmitConfidential: refuse to submit a cross-chain bulk without a fee estimate (new missingNetworkFee i18n string in all 10 locales)

Regression tests: bulk-fee.test.ts (gating, estimator-throw flagging, the pre-fix source-network failure mode) and lib/intents-fee.test.ts (mismatch throws). bun test green; tsc/biome introduce no new diagnostics on touched files.

Claude-Session: https://claude.ai/code/session_018PUYhFCYqLcQEDbcVqebMH


frol:
@denbite tested it and said that it does not work as expected, so he will author a new PR

PR #1054, opened 2026-07-07
Job resilience: retry + catch_panic + supervised restart + failure aggregation

frol-ai:
Reviews #1043 ("Make all apalis jobs recover from errors and keep running") for applicability now that the apalis jobs (#1042) + follow-up fixes (#1048/#1050) have merged to `main`, ports what's still relevant, and adds failure aggregation. Base = `main`.

### Is #1043 still applicable?
Partly — `main` already carries some of it. Assessment vs current `main`:
- **Ported (still needed):** worker containment — `catch_panic` + `Monitor`-supervised restart + apalis's Sentry layer — and multi-step **failure aggregation** in handlers.
- **Already in `main`:** reworked `schedule_every_secs` (incl. the ≥60-min "runs hourly" bug fix), retention-days clamp, and the apalis migration-collision fix (`setup_apalis`).
- **Deliberately NOT taken:** (1) #1043's blanket per-task **`RetryPolicy`** — several handlers have non-idempotent side effects (`bulk_payment_payout`'s on-chain `payout_batch`, `ft_lockup`'s `claim` txs, Telegram sends), so retrying the whole task risks duplicate transactions; the cron schedule is the retry, and apalis-postgres already retries transient backend errors under `run()`. (2) #1043 reverts `price_sync` to a non-erroring `String` return — a prior review fix, so `price_sync` stays `Result<String, _>` (failed runs stay failed on the board).

### Worker containment
Each cron worker is registered on an apalis `Monitor` via a factory and wrapped with:
1. **`catch_panic`** — a panicking handler becomes a task error instead of killing the worker (adds the apalis `catch-panic` feature).
2. **`Monitor` supervision** — `should_restart` rebuilds and restarts a worker that exits. Restart is **immediate** (the hook is synchronous, so it can't sleep), which does **not** hot-loop: apalis-postgres's fetcher backs off internally (1s→5min) inside `run()`, so a worker only exits on a terminal condition, not on transient DB blips.
3. **Sentry** — apalis's `SentryLayer` reports task failures to the shared Sentry hub; `TraceLayer` logs failures at WARN so we don't double-report.
4. **No task-level retry** (see "Deliberately NOT taken") — non-idempotent handlers must not re-run automatically.

Graceful shutdown drains in-flight tasks on **SIGINT or SIGTERM** (SIGTERM is what containers/systemd send); a failed signal-handler install logs and leaves the monitor running rather than shutting it down.

### Failure aggregation
Multi-step handlers now run every step and aggregate failures instead of aborting on the first `?`, returning the **original** error value (not a stringified copy) to preserve the source chain for Sentry grouping:
- `notifications` (detect + dispatch) — a bad Telegram token no longer stalls detection.
- `subscription_monthly_reset` (credit reset + export expiry).
- `ft_lockup_refresh` (schedule refresh + due claims).
- `goldsky_enrichment` surfaces partial progress on a mid-drain failure (cursor persisted per batch; next tick resumes).

### More stability considered, skipped
A blanket per-task **timeout** (apalis `timeout` feature) was evaluated and **skipped**: heavy jobs (public-dashboard refresh, confidential snapshots/reconciliation) can legitimately run long, so a global timeout would fail real work. `catch_panic` + supervised restart already cover the crash/exit modes; per-queue timeouts can be added later once real runtimes are known.

`cargo clippy -- -D warnings` clean; `cargo fmt --check` clean; jobs unit tests pass (incl. the migration-coexistence regression test).

### DIFF
--- a/nt-be/Cargo.toml
+++ b/nt-be/Cargo.toml
@@ -56,7 +56,7 @@ bs58 = "0.5"
 url = "2"
 teloxide = { version = "0.17.0", features = ["rustls"], default-features = false }
 # Background job processing (apalis + web board)
-apalis = { version = "1.0.0-rc.9", features = ["limit", "retry", "tracing"] }
+apalis = { version = "1.0.0-rc.9", features = ["limit", "retry", "tracing", "catch-panic"] }
 apalis-core = { version = "1.0.0-rc.9", features = ["serde"] }
 apalis-cron = { version = "1.0.0-rc.8", features = ["serde"] }
 apalis-sql = "1.0.0-rc.8"
--- a/nt-be/docs/BACKGROUND_JOBS.md
+++ b/nt-be/docs/BACKGROUND_JOBS.md
@@ -15,6 +15,30 @@ What this replaces: 17 hand-rolled `tokio::spawn` + interval loops in
 - per-task tracing spans; `concurrency(1)` guarantees cycles never overlap
 - failed cycles are visible as failed tasks instead of just log lines
 
+## Resilience
+
+Every worker is wrapped in three layers of failure containment
+(`spawn_cron_worker!` in `src/jobs/mod.rs`), so a job keeps running through
+transient faults instead of dying until the next process restart:
+
+1. **Task errors** — a failing task is retried up to `TASK_RETRIES` (2)
+   times; after that the failure is recorded and the next cron tick starts
+   a fresh task. One bad cycle never stops the schedule.
+2. **Handler panics** — `catch_panic` turns a panic into a task error
+   (same path as #1) instead of tearing the worker down.
+3. **Worker-loop exit** — on a sustained backend/storage failure the
+   supervisor rebuilds and restarts the worker with capped exponential
+   backoff (1s → 60s), forever; the backoff resets after a run that stayed
+   up `RESTART_HEALTHY_AFTER` (5 min). A worker never dies silently.
+
+Handlers that do several steps run **all** steps and aggregate failures
+rather than aborting on the first error (e.g. notifications detect +
+dispatch, monthly credit reset + export expiry, FT-lockup refresh +
+claims); the goldsky drain reports partial progress on a mid-drain failure
+(its cursor is persisted per batch). Startup migrations for apalis are
+tracked in a private `apalis_migrations` schema so they never collide with
+the app's own `_sqlx_migrations`.
+
 ## Web UI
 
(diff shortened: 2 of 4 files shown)


frol-ai:
Pushed a follow-up (`36fb4d26`) replacing the hand-rolled per-worker `tokio::spawn` + restart/backoff loop with apalis's own **`Monitor`** — which is exactly what that manual glue was reinventing.

- `spawn_all` now builds one `Monitor`, registers every worker via a factory (Monitor rebuilds the worker on restart with a fresh connection), and drives them on a single supervised task.
- `should_restart` restarts a worker that exits on a backend/storage failure; a **clean exit on shutdown is honoured**, so in-flight tasks drain instead of being fought by a restart (the old loop restarted on clean exit too — fighting graceful shutdown).
- `run_with_signal(ctrl_c)` adds coordinated graceful shutdown we didn't have.

Task-level resilience is unchanged (retry + catch_panic) and paced by the cron schedule, so dropping the custom exponential backoff doesn't reintroduce hot-looping — a worker only exits on a rare terminal backend failure. If a sustained-outage cooldown is wanted later, apalis's native `circuit_breaker` ext (`recovery_timeout`) is the tool.

clippy -D warnings + fmt clean; jobs unit tests pass.

frol-ai:
Reviewed the Sentry integration and wired apalis's built-in one (`9d369d65`).

**Current state:** the app inits Sentry in `observability.rs` and bridges `tracing → Sentry` (ERROR → event). apalis's tracing layer already logs task failures at ERROR, so job failures *did* reach Sentry — but only as generic log events (message-string grouping, no queue/attempt tags, no APM).

**Change:** adopt apalis's dedicated `SentryLayer` (feature `sentry`). It shares the same `sentry-core` 0.48.2 as the app (verified via `cargo tree`), so it uses the already-initialised global hub — no second client. Per task it:
- opens a Sentry **performance transaction** (APM), and
- on failure `capture_error`s the real error with **queue / task-id / attempt** context → proper grouping + filtering.

Layered **outside** `retry`/`catch_panic`, so a failure is captured **once** (after all retries, panics included). To avoid the tracing bridge double-reporting the same failure, the per-task tracing layer now logs failures at **WARN** (a single retried cycle is a warning; persistent failures still surface via Sentry + the board). Worker-level errors (exit/restart) stay at ERROR. Fully no-op when `SENTRY_DSN` is unset.

clippy -D warnings + fmt clean; jobs unit tests pass.

Issue #838, opened 2026-06-09
[BUG] Trezu - Request Details: Historical Requests use live Market Price instead of Historical Execution Price


rknearn-qa:
### Summary

Request Details display the **current market price** of a token instead of the token price at the moment the request was executed.

As a result, historical requests show changing USD conversion rates over time, causing inaccurate financial records and inconsistencies between the original transaction value and the value currently displayed in Trezu.

### Preconditions:

- Payment or Exchange request has been executed (https://testenv.trezu.app/staging-qa.sputnik-dao.near/requests/12)
- Token price has changed since execution
- Request Details page is available

### Steps To Reproduce

1. Execute a Payment or Exchange request
2. Wait until the token market price changes
3. Open the executed request in **Request Details**
4. Observe the displayed token price / USD conversion rate
5. Compare it with the market price at the time the request was executed

### Actual Result:

- Request Details show the **current market price**
- Historical requests display different USD values over time
- Financial information changes despite the transaction being completed

Example:

```
Request executed:
1 NEAR = $1.60

Current market price:
1 NEAR = $3.00

Request Details display:
1 NEAR = $3.00
```

<img width="1361" height="691" alt="Image" src="https://github.com/user-attachments/assets/9c0cf4d8-3dda-4a17-bbb5-5c06af10cfc2" />
<img width="1362" height="679" alt="Image" src="https://github.com/user-attachments/assets/9067ec7f-f52d-4894-b029-3b36c918b4ff" />

### Expected Result:

- Request Details should display the token price that existed when the request was executed
- Historical requests should remain immutable
- USD values should not change after execution

Example:

```
Request executed:
1 NEAR = $1.60

Current market price:
1 NEAR = $3.00

Request Details should display:
1 NEAR = $1.60
```

### Special Notes:

```
Possible Root Cause:

* Request Details fetch current token prices from the pricing service
* Historical execution price is not stored with the transaction
* UI recalculates USD values using live market data
* Request metadata lacks a snapshot of execution-time exchange rates
* Same issue may affect PDF receipts and exported reports
```

### Workarounds:

- Use external historical price providers to determine the asset price at the execution timestamp

### Repro Rate:

Always (10/10)

### Setup - OS - Browser - Environment

* OS / Devices: Windows 11 (x64), Motorola G23 (Android 14), Iphone XR (iOs 18.7.6)
* Screen resolution(s): 1366 x 768, 1600 x 720, 1792×828
[...]

Issue #663, opened 2026-05-14
[Task] Retroactively Update Historical USD Values for Past Transactions


ori-near:
### Background

After https://github.com/NEAR-DevHub/trezu/issues/662 is complete, we need to update existing historical transactions to show the USD values at the time of the execution.

### User Story

As a user, I want all my historical transaction data to be updated with the correct USD equivalent value at the time of execution so my entire transaction history is accurate for auditing.

### Acceptance Criteria

- [ ] Identify a suitable data provider for historical token prices. Backfill the correct historical USD price for all past transactions. Then display them in the app in the appropriate sections.

rknearn-qa:
Implemented

<img width="1252" height="683" alt="Image" src="https://github.com/user-attachments/assets/290c93e1-8df1-4dbf-ae6c-f4c07e163b1c" />
<img width="1365" height="718" alt="Image" src="https://github.com/user-attachments/assets/51176394-2249-4403-8b07-b8d6bad797e7" />

* OS / Devices: Windows 11 (x64), Motorola G23 (Android 14), Iphone XR (iOs 18.7.6)
* Screen resolution(s): 1366 x 768, 1600 x 720, 1792×828
* Browser(s) + version(s): Chrome (150.0.7871.129), Mozilla Firefox (151.0.3), Brave (1.91.168), Safari (18.7.4), Edge (149.0.4022.52)
* Wallet: Meteor
* App environment: Staging (trezu.app)
* Near Network: Mainnet

PR #1066, opened 2026-07-09
added: usd backfill

ktarun1419:
#663 #838

- Fixed unique key constraint error  (The enqueue logic treats `failed` jobs with remaining attempts as still active, but the unique inflight index only protects `Pending/Queued/Running`. That allows a duplicate job with the same `job_key` to be inserted while the original failed job is still retryable)


frol (approved):

frol:
It feels almost too complex for the job it does, but we could discuss further refactoring separately, it is not a blocker for the PR.

PR #1099, opened 2026-07-13
bulk activation: fix the subaccount auth signing path (JWT was never issued)

frol-ai:
## Root cause (confirmed: `bulk_payment_access_token` NULL even after approval)
The bulk-payment subaccount's 1Click authentication was signing with the **wrong key**, so 1Click rejected it and no JWT was ever issued/stored — on **both** paths (creation-time provisioning and the manual activation proposal). That's why freshly created *and* approved-activation treasuries stayed blocked.

`bootstrap` registers the DAO's own MPC key — `derive(dao, dao_id)` — under the `<prefix>.bulkpayment.near` subaccount on intents.near (the same key the DAO uses for its own confidential auth). But:
- `build_bulk_payment_auth_proposal` signed the `v1.signer` request with `path = ""` → `derive(dao, "")`, a key **not** registered for the sub.
- the vote relay fetched the empty-path MPC key to hand to 1Click for `bulk_auth`.

So the signature/public-key didn't match what intents.near/1Click expected → auth failed → the token-store `if let Some(accessToken)…` was skipped → column stayed NULL.

## Fix
Sign the bulk auth with `path = dao_id` (the JWT is still issued for the sub via the auth message's `signer_id`), and have the relay fetch the DAO-path MPC key for `bulk_auth` too. This matches the flow already documented in `ensure_bulk_subaccount` ("path=<dao_id> … valid for <sub> because bootstrap registered the DAO's pubkey under <sub>").

Touches `build_bulk_payment_auth_proposal` + its two callers (prepare, creation setup) and the relay's `mpc_path`.

## Note
Existing treasuries stuck with a NULL token just need to re-run activation (Start activation → approve) after this ships — the retry now authenticates with the correct key. Complements #1097 (which stopped the status endpoint from falsely blocking the form).

## Verification
`cargo clippy --lib --bins -- -D warnings` + `cargo fmt` clean. (End-to-end auth is validated by the bulk-payment E2E, which exercises the sandbox activation path.)

### DIFF
--- a/nt-be/src/handlers/intents/confidential/bulk_activation.rs
+++ b/nt-be/src/handlers/intents/confidential/bulk_activation.rs
@@ -261,7 +261,7 @@ pub async fn prepare_bulk_activation(
 
     // 2. Build the auth proposal the multisig must approve.
     let (proposal, auth_payload) =
-        build_bulk_payment_auth_proposal(&state, sub_id.as_str()).await?;
+        build_bulk_payment_auth_proposal(&state, sub_id.as_str(), dao_id.as_str()).await?;
     let payload_hash = compute_nep413_hash(&auth_payload).ok_or_else(|| {
         (
             StatusCode::INTERNAL_SERVER_ERROR,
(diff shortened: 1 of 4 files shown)


frol:
On the second try, the bulk activation worked fine. I am not sure what that was, so we will need to keep an eye on it.

Issue #1101, opened 2026-07-14
[BUG] AURORA token displays omni in chart and payments


Megha-Dev-19:
AURORA token display omni in chart/payments but in dashboard assets it's just AURORA. 
Make the naming consistent across the app

rknearn-qa:
Retested
Fixed

<img width="1364" height="677" alt="Image" src="https://github.com/user-attachments/assets/25f800d4-4549-4f21-959e-1e13f6e0199d" />
<img width="1343" height="683" alt="Image" src="https://github.com/user-attachments/assets/f96dd205-b52d-4618-ba0a-632dcbb2d0de" />
<img width="1040" height="667" alt="Image" src="https://github.com/user-attachments/assets/b591cac1-78ae-4441-a0c5-a4fbd1b0c544" />
<img width="850" height="640" alt="Image" src="https://github.com/user-attachments/assets/bbf17070-3d65-4e88-b094-e8e02140fffe" />

* OS / Devices: Windows 11 (x64), Motorola G23 (Android 14), Iphone XR (iOs 18.7.6)
* Screen resolution(s): 1366 x 768, 1600 x 720, 1792×828
* Browser(s) + version(s): Chrome (149.0.7827.54), Mozilla Firefox (151.0.3), Brave (1.91.168), Safari (18.7.4), Edge (149.0.4022.52)
* Wallet: Meteor
* App environment: Staging (trezu.app)
* Near Network: Mainnet

Issue #954, opened 2026-06-22
[BUG] Trezu - Executed Requests & PDF Receipts: Execution Timestamp initially displays "N/A" until page refresh


rknearn-qa:
### Summary

Immediately after a request is approved and executed, the execution timestamp is not available in both the **Executed Requests** list and the generated **PDF Receipt**.

Instead of displaying the actual execution time, the UI shows **"N/A"**. After refreshing the page, the correct timestamp appears.

This indicates a synchronization or data-refresh issue where execution metadata is not immediately available after request completion. Relates to #952 

### Preconditions:

* A pending request exists
* User has permission to approve requests

### Steps To Reproduce

1. Open a pending request
2. Approve the request
3. Navigate to **Executed Requests**

Example:

```text
https://testenv.trezu.app/confidential-staging.sputnik-dao.near/requests?tab=Approved
```

4. Locate the newly executed request
5. Observe the execution timestamp
6. Generate a PDF Receipt for the same request
7. Observe the execution timestamp in the PDF
8. Refresh the page
9. Observe the timestamp again

### Actual Result:

Immediately after execution:

```text
Execution Time: N/A
```

* Executed Requests page displays `N/A`
* PDF Receipt displays `N/A`

After refreshing the page:

* Correct execution timestamp is displayed
* PDF Receipt contains the correct timestamp

https://github.com/user-attachments/assets/311dbfc4-ec6d-429e-9898-489e297ed65d

### Expected Result:

The execution timestamp should be available immediately after request execution without requiring a page refresh.

Example:

```text
Executed At:
Jun 22, 2026, 3:45 PM UTC+3
```

### Special Notes:

```text
Possible Root Cause:

* Execution metadata is not refreshed after approval completes
* UI renders before execution timestamp is available
* Cache/state is not invalidated after successful execution
* PDF generation uses stale request data immediately after execution
* Backend updates execution timestamp asynchronously while frontend expects immediate availability
```

### Workarounds:

* Refresh the page
* Reopen the request details page
* Generate the PDF again after the page reloads

### Repro Rate:
Always (10/10)

### Setup - OS - Browser - Environment

* OS / Devices: Windows 11 (x64), Motorola G23 (Android 14), iPhone XR (iOS 18.7.6)
* Screen resolution(s): 1366 × 768, 1600 × 720, 1792 × 828
[...]

PR #1102, opened 2026-07-14
Fix AURORA symbol override and post-approval execution timestamp N/A

Megha-Dev-19:
- Prefer the unified AURORA display symbol/name over the standalone `AURORA (omni)` catalog duplicate in token metadata. #1101 
- After a request is executed, poll NearBlocks for the execution tx and keep a loading state instead of showing "N/A" until the timestamp resolves. #954

### DIFF
--- a/nt-fe/features/proposals/components/proposals-table.tsx
+++ b/nt-fe/features/proposals/components/proposals-table.tsx
@@ -67,9 +67,7 @@ import { useProposalsInsufficientBalance } from "../hooks/use-proposals-insuffic
 import { useProposalTransaction, useSwapStatus } from "@/hooks/use-proposals";
 import {
     extractReceiptProposalData,
-    getProposalExecutedDate,
-    isExecutionTimestampPending,
-    isTerminalSwapStatus,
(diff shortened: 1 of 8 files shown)


nt-fe/features/proposals/components/proposals-table.tsx
```diff
@@ -132,14 +137,24 @@ function ProposalTimelineDate({
...
+        !isTerminalSwapStatus(swapStatus?.status);
+    const isDateLoading = isExecutionTimestampPending({
+        executedDate,
+        isQueryLoading: shouldUseSwapDate
+            ? isLoadingSwapStatus
+            : isLoadingTransaction,
+        isAwaitingResolution: shouldUseSwapDate
+            ? isAwaitingSwapDate
+            : isAwaitingTransaction,
+    });
```
frol: Is there a way to reduce code duplication between proposals-table and proposal-sidebar and receipt/page? I fear that these conditions may diverge from each other one day...

frol (approved):
My comment is not a blocker, so feel free to merge this PR once you are ready

Issue #1070, opened 2026-07-09
Add Google Tag Manager for marketing conversion tracking


Megha-Dev-19:
Joe mentioned to add Google Tag Manager (GTM) to track conversions and manage ad/analytics tags.

rknearn-qa:
Implemented

<img width="1365" height="702" alt="Image" src="https://github.com/user-attachments/assets/d824b294-dcca-45e7-a24a-8e5ae6e81b89" />

* OS / Devices: Windows 11 (x64), Motorola G23 (Android 14), Iphone XR (iOs 18.7.6)
* Screen resolution(s): 1366 x 768, 1600 x 720, 1792×828
* Browser(s) + version(s): Chrome (149.0.7827.54), Mozilla Firefox (151.0.3), Brave (1.91.168), Safari (18.7.4), Edge (149.0.4022.52)
* Wallet: Meteor
* App environment: Production (trezu.app)
* Near Network: Mainnet

PR #1104, opened 2026-07-14
feat: add Google Tag Manager for conversion tracking

Megha-Dev-19:
#1070


frol (approved):
@Megha-Dev-19 Please, configure the `NEXT_PUBLIC_GTM_ID` for production ~and staging~ right away, so we don't forget to set it to the proper value.

PR #1105, opened 2026-07-14
fix(bulk-payment): use `EXACT_OUTPUT` quotes so recipients receive the full amount

denbite:
This PR addresses transfer fees silently eating into bulk payment amounts, so recipients received less than the DAO approved.

### Example

DAO bulk-pays 5 USDC to each of two Solana recipients, where each cross-chain transfer costs ~0.4 USDC in fees.

#### Before

With `EXACT_INPUT`, fees were deducted from the transfer amount: the DAO paid 10, but each recipient received `5 − 0.4 = 4.6` — 8% less than the approved amount.

#### After

With `EXACT_OUTPUT`, fees are added on top: each leg quotes `amountIn = 5.4`, the header funds the subaccount with exactly `5.4 + 5.4 = 10.8`, the DAO pays 10.8, and each recipient receives exactly 5.

### DIFF
--- a/nt-be/src/handlers/intents/confidential/bulk_payment_prepare.rs
+++ b/nt-be/src/handlers/intents/confidential/bulk_payment_prepare.rs
@@ -106,7 +106,7 @@ fn build_recipient_quote_body(
 
     let mut body = json!({
         "dry": false,
-        "swapType": "EXACT_INPUT",
+        "swapType": "EXACT_OUTPUT",
         "slippageTolerance": slippage,
         "originAsset": origin_asset,
         "depositType": "CONFIDENTIAL_INTENTS",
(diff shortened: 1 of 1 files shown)


frol (changes requested):
This is not going to work nicely. We deliberately used EXACT_INPUT here, and we need to do best effort of adding the fee on top of it manually.

Since we first transfer the amount to the bulkpayment subaccount, and approval may take several days, there is a risk that we will transfer 10 USDC to the subaccount and EXACT_OUTPUT would require to have 10.8 USDC or 10.9 USDC on the balance to complete all the transfers.

denbite:
@frol Notice that the header quote is also built with `EXACT_OUTPUT`, and its `amount` is the sum of the recipient quotes' `amountIn` (which already include the fees). So the subaccount is guaranteed to receive exactly what the recipient legs need - in your example it would be funded with 10.8, not 10, and the scenario of the subaccount balance falling short cannot occur

The DAO side is covered the same way - the amount pulled from the DAO comes from the header quote's own `amountIn`, so fees are allocated on top at every level rather than deducted

frol:
As we discussed offline, the AmountIn in the quote with EXACT_OUTPUT intent is not the guaranteed amount and can still be adjusted by NEAR Intents backend, so we have to use EXACT_INPUT and estimate the fee and embrace the fact that the recipients may receive slightly more or less amount (hopefully within $0.01 range)

frol:
Superseded by #1117

PR #1106, opened 2026-07-14
selectmodal bug fixed

### LINKED ISSUE #1098: [BUG] Trezu - Deposit: Asset selector briefly reopens after double-clicking an asset
### Summary

When selecting an asset in the **Deposit** flow, rapidly double-clicking an asset can cause the **Asset Selector** modal to briefly reopen before closing again.

The issue is intermittent but reproducible under certain conditions, particularly when selecting an asset for the first time after opening the page or choosing an asset that was not previously selected.

### Preconditions

* Public or Confidential Treasury exists.

### Steps To Reproduce

1. Open **Deposit**.
2. Click the **Asset** selector.
3. Rapidly double-click one of the popular assets.
4. Observe the modal behavior.

### Actual Result

* The selected asset is applied.
* The Asset Selector modal briefly appears again (less than one second) before closing.
* A noticeable UI flicker/glitch occurs.

https://github.com/user-attachments/assets/b1ada470-8401-4a0e-844b-c077bdb742ab

### Expected Result

* A single click (or accidental double-click) should result in exactly one selection.
* The Asset Selector should close smoothly once.
* No additional modal opening or flickering should occur.

### Special Notes

```
Observed Behavior:

The issue is intermittent and occurs more frequently:

• On the first asset selection after opening Deposit.
• When selecting an asset that has not been selected previously.

Possible Root Cause:

A race condition or duplicate click event causes the modal state to
toggle twice.

Potential areas to investigate:

- Double invocation of the asset selection handler.
- Event bubbling (onClick/onPointerUp).
- Modal open/close state updates.
- React state updates after asset selection.
- Missing debounce/throttle or click guard.
```

### Workarounds

* Select assets with a single click.

### Repro Rate
Intermittent (timing dependent).

### Setup - OS - Browser - Environment

* OS / Devices: Windows 11 (x64), Motorola G23 (Android 14), Iphone XR (iOs 18.7.6)
* Screen resolution(s): 1366 x 768, 1600 x 720, 1792×828
* Browser(s) + version(s): Chrome (149.0.7827.54), Mozilla Firefox (151.0.3), Brave (1.91.168), Safari (18.7.4), Edge (149.0.4022.52)
* Wallet: Meteor
* App environment: Staging (trezu.app)
* Near Network: Mainnet

### Impact

**User Impact**

* Creates a visible UI flicker.
* Makes asset selection feel unstable or unresponsive.
* May lead users to believe their selection did not register correctly.

**System Impact**

* Indicates duplicate state transitions or event handling in the Asset Selector.
* Although functionality is preserved, the modal lifecycle is inconsistent.

**Business / Release Risk**

* Cosmetic/UX issue with no apparent data loss or incorrect behavior.
* Should be addressed to improve interaction quality.

### QA Verdict

**Verdict:** Lowest issue

**Rationale:**

While the selected asset is ultimately applied correctly, the modal should not reopen after a completed selection. This behavior indicates an unnecessary UI state transition and degrades the user experience.

**Recommendation**

* Prevent duplicate selection events while the modal is closing.
* Debounce or ignore repeated clicks on asset items.
* Review modal open/close state updates after asset selection.
* Add regression coverage for:

  * Rapid double-clicks.
  * First asset selection after page load.
  * Previously unselected assets.
  * Slow rendering/network conditions.
  * Mouse and touch interactions.

### Priority

⚪ P4 : Low

himanshuranjan007:
fixes #1098 

two changes made in select-modal 
useRef(false) flag that is set to true on first invocation , immediately any subsequent calls while the flag is set to true 
& useEffect resets to false whenever isOpen becomes true

https://github.com/user-attachments/assets/9e9611c4-cd03-45c9-b696-02c97ee9991f


frol:
I feel that we might need to look deeper into the problem as I don't want to fix symptoms instead of the root cause. Also, the issue itself is quite odd in the first place, so I don't want to bring a symptomatic hotfix for the issue that only occurs if the user does some odd things.

As pointed out by Copilot review, this hotfix did not account for the multiselect case and as such this fix for a corner case would break a multiselect case, which is a much more severe problem, and who knows what else would it break.

frol (changes requested):
See my comment and the Copilot review as well

PR #1117, opened 2026-07-15
fix: pull transfer fees from actual quotes to render on "Confirm Confidential Bulk Payment" page on UI

denbite:


nt-be/src/handlers/intents/confidential/bulk_payment_confirm.rs
```diff
@@ -0,0 +1,495 @@
...
+    pub notes: Option<String>,
+}
+
+#[derive(Serialize, Debug)]
+#[serde(rename_all = "camelCase")]
+pub struct BulkPaymentConfirmResponse {
+    pub success: bool,
+}
+
+pub async fn bulk_payment_confirm(
```
frol, line 39: It seems that if the frontend would skip calling the "confirm" endpoint, they won't be able to set the notes, but otherwise they can proceed without spending the bulk-payment credit. Do I read it right? We could add a new `status` (confirmed?) for `confidential_bulk_payments` table, but it might be not worth the trouble. Still, I just wanted to make sure that we are on the same page here, and I did not miss anything.

frol (changes requested):
More general comment is that it feels like on the frontend we add something quite specific to only confidential bulk payments, but I feel like the network fees calculation should be mostly the same for both public and confidential bulk payments, so it would be ideal to generalize and align the implementations, so they are agnostic of whether it is a public or confidential treasury, unless there is a good reason not to (please, feel free to push back)

frol (approved):

frol:
We discussed all the topic in with Den in DM and decided to go as-is

Issue #1116, opened 2026-07-15
[BUG] Trezu - Manage Treasuries: Hidden Treasury displays Guest badge and Save button for Treasury Members


rknearn-qa:
### Summary

When a treasury is hidden from the **Manage Treasuries** page, opening it via **View Treasury** incorrectly displays the **Guest** badge and **Save** button, even though the user is a member of that treasury.

The membership state is incorrectly resolved after navigating from a hidden treasury, causing member UI to fall back to the guest experience.

### Preconditions

* Treasury exists.
* User is a member of the treasury.
* Treasury is visible in **Manage Treasuries**.

### Steps To Reproduce

1. Open **Manage Treasuries**.

   ```
   https://testenv.trezu.app/app/manage-treasuries
   ```
2. Hide one of your member treasuries.
3. Click **View Treasury** for the hidden treasury.
4. Observe the sidebar and header UI.

### Actual Result

* **Guest** badge is displayed.
* **Save** button is displayed.
* User is incorrectly treated as a guest despite being a treasury member.

<img width="1365" height="689" alt="Image" src="https://github.com/user-attachments/assets/84d88764-9bb8-4947-b5bd-357689ab6df5" />

https://github.com/user-attachments/assets/187baa82-56bc-483a-a955-7a2728ff24d7

### Expected Result

* Hidden treasuries should still recognize the user's membership.
* Member UI should be displayed.
* Guest-specific controls (Guest badge, Save button, etc.) should not appear.

### Special Notes

```
Possible Root Cause:

Hidden treasuries appear to be excluded from the membership lookup after
navigation, causing the frontend to classify the user as a guest.

The visibility (hidden) state should not affect membership resolution.
```

### Workarounds
None

### Repro Rate
**Always (10/10)

### Setup - OS - Browser - Environment

* OS / Devices: Windows 11 (x64), Motorola G23 (Android 14), Iphone XR (iOs 18.7.6)
* Screen resolution(s): 1366 x 768, 1600 x 720, 1792×828
* Browser(s) + version(s): Chrome (149.0.7827.54), Mozilla Firefox (151.0.3), Brave (1.91.168), Safari (18.7.4), Edge (149.0.4022.52)
* Wallet: Meteor
* App environment: Staging (trezu.app)
* Near Network: Mainnet

### User Impact

* Members are incorrectly presented with the guest experience.
* Creates confusion regarding access rights.
* Member-specific navigation and actions may appear unavailable.

### System Impact

[...]

Issue #1108, opened 2026-07-14
[DEV] Add ability to enter amount in the Receive field during exchange


olskik12:
**Context**
Currently, the user can only enter an amount in the Sell field (the amount they want to send). There is no option to specify the desired amount to receive.

<img width="1454" height="835" alt="Image" src="https://github.com/user-attachments/assets/ef86bd03-87c4-4be2-8a1e-31b5d1e8d8a0" />

**Expected behavior**
As a user, when entering an amount in the exchange input, I expect to be able to enter either the amount I want to send (Sell) or the amount I want to receive (Receive).

**What needs to be done**
Give users the ability to edit the Receive field. In this case, the Sell field should automatically recalculate and display the amount the user needs to send in order to receive the specified amount in Receive.

**Acceptance Criteria**

- [ ] User can enter an amount in the Sell field → Receive is recalculated automatically.
- [ ] User can enter an amount in the Receive field → Sell is recalculated automatically.
- [ ] Recalculation is based on the current exchange rate.

rknearn-qa:
Implemented

<img width="1365" height="688" alt="Image" src="https://github.com/user-attachments/assets/978b8590-16a1-4100-8b5d-304b235251df" />

* OS / Devices: Windows 11 (x64), Motorola G23 (Android 14), Iphone XR (iOs 18.7.6)
* Screen resolution(s): 1366 x 768, 1600 x 720, 1792×828
* Browser(s) + version(s): Chrome (149.0.7827.54), Mozilla Firefox (151.0.3), Brave (1.91.168), Safari (18.7.4), Edge (149.0.4022.52)
* Wallet: Meteor
* App environment: Staging (trezu.app)
* Near Network: Mainnet

PR #1120, opened 2026-07-16
feat: allow exact-output amounts on exchange

Megha-Dev-19:
#1116 
#1108


nt-fe/components/token-input.tsx
```diff
@@ -195,161 +201,262 @@ export function TokenInput<
...
+                                    className="text-sm mt-2"
+                                />
+                            ) : infoMessage ? (
+                                <p className="text-general-info-foreground text-sm mt-2">
+                                    {infoMessage}
+                                </p>
+                            ) : !balanceWarning ? (
+                                <p className="text-muted-foreground text-xs invisible">
+                                    Invisible
+                                </p>
```
frol, line 430: This is not a new change, but this caught my eye. I don't understand what is going on here. Do you?

nt-fe/components/token-input.tsx
```diff
@@ -195,161 +201,262 @@ export function TokenInput<
...
+                                            })}
+                                        </p>
+                                        {!readOnly && (
+                                            <Button
+                                                type="button"
+                                                variant="secondary"
+                                                className="bg-muted-foreground/10 hover:bg-muted-foreground/20"
+                                                size="sm"
+                                                onClick={() => {
+                                                    if (
```
frol: It is really hard to review these inline handlers (there are 3 more below). Can they be extracted into variables above the JSX syntax?

frol (approved):

nt-fe/components/token-input.tsx
Megha-Dev-19: When there’s no error, warning, or info message, it still renders a hidden <p> (invisible = takes space, not shown) so the token input block doesn’t jump in height when a message appears or disappears.

Issue #1107, opened 2026-07-14
[DEV] Update chart endpoint to use new confidential/public treasury infra


Megha-Dev-19:
Update the chart endpoint to use the new infra used for confidential and public treasuries.
Right now chart data can be inconsistent with Recent Activity. After this change, chart should reflect correct balances/history for both treasury types.

Issue #957, opened 2026-06-22
[BUG] Incorrect “To” and “From” Addresses in Recent Transactions


olskik12:
Background

As a user, I want to transfer USDC/Solana tokens from one treasury to another.

After receiving the destination address and completing the payment request, I noticed an issue in the transaction details:
Current Behavior
1. In the source treasury (the treasury that sent the payment), the Recent Transactions details display an incorrect recipient address instead of the address specified in the payment request.
2. In the destination treasury, the transaction shows an incorrect sender.

Expected Behavior
1. In the source treasury, the transaction details should display the recipient address from the payment request.
2. In the destination treasury, the transaction details should display the sender address.

https://github.com/user-attachments/assets/7afd54bb-e056-4929-a5b6-3fcaf0e809a9

the source treasury

<img width="1655" height="885" alt="Image" src="https://github.com/user-attachments/assets/44824e1a-bd17-4b59-8e7a-af3e2c4c34ee" />

The destination treasury

<img width="1656" height="890" alt="Image" src="https://github.com/user-attachments/assets/f177919f-7214-4576-9c9e-cdab87b30a83" />

olskik12:
I created this ticket as a reminder of how this should work once @ktarun1419  improves the backend for Recent Transactions in the public treasury

cc @rknearn-qa

rknearn-qa:
Retested
Still reproducing

<img width="1365" height="691" alt="Image" src="https://github.com/user-attachments/assets/908c7e68-225d-4891-ad10-f20a668a0ba1" />
<img width="1365" height="689" alt="Image" src="https://github.com/user-attachments/assets/434a3d2a-7a5c-4588-b41b-948a510f21b0" />
<img width="1365" height="688" alt="Image" src="https://github.com/user-attachments/assets/51230c68-e1aa-4567-8d11-5645759850c7" />

* OS / Devices: Windows 11 (x64), Motorola G23 (Android 14), Iphone XR (iOs 18.7.6)
* Screen resolution(s): 1366 x 768, 1600 x 720, 1792×828
* Browser(s) + version(s): Chrome (150.0.7871.129), Mozilla Firefox (151.0.3), Brave (1.91.168), Safari (18.7.4), Edge (149.0.4022.52)
* Wallet: Meteor
* App environment: Staging (trezu.app)
* Near Network: Mainnet

Issue #616, opened 2026-05-07
[Dev] Recent Transactions — Improvement


olskik12:
Implement the improved recent transactions flow and make it consistent between private and public treasuries.
the ticked - #405 

Criteria
- Do not display “Earning” (staking rewards) transactions on the dashboard in the recent transaction section.
- Each transaction should have a status indicating whether it is completed or still in progress.
- Exchange transactions should be displayed as a single “Exchange” transaction.
- If the transaction was executed from the treasury (not a deposit), the user should be able to navigate to the request.
- On older DAOs (created on Near Treasury), users will also see linked transactions (Proposal Bond)

rknearn-qa:
Almost all sub-tickets are closed
Move it to Done

PR #1122, opened 2026-07-16
chore: history UI improvements

ktarun1419:
#616 #1107 #1123 #838 #1124 #957

- Reduced public history worker concurrency to lower database pool pressure during ingestion, backfill, and projection jobs.
- Added synthetic quote-pending history rows from submitted/executed proposals so the UI can show pending payments/exchanges before the real transfer is finalized.
- Updated activity UI to render pending/failed public payments and exchanges more clearly.
- Fixed public history backfill by resetting affected cursors after the NearBlocks pagination issue.
- Added quote-status refresh/projection support so pending quote rows can resolve into successful, failed, or refunded history entries.
- Improved receipt/rate rendering to use proposal quote USD values as fallback.


nt-be/src/handlers/balance_changes/history.rs
```diff
@@ -1578,6 +1578,10 @@ pub struct SwapInfo {
     pub solver_transaction_hash: String,
     /// "deposit" for the outgoing leg, "fulfillment" for the incoming leg
     pub swap_role: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub sent_amount_usd: Option<f64>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub received_amount_usd: Option<f64>,
```
frol, line 1584: Do we use f64 for all _usd estimated amounts? I am a bit worried about it, but since we only show it as an estimate, maybe it is fine, though I would prefer decimals or fixed decimals (e.g. store values in integer cents).

P.S. I am open for a push back

nt-be/src/routes/balance_changes.rs
```diff
@@ -163,263 +171,267 @@ pub async fn get_balance_changes_internal(
...
-        }
-    }
-
-    Ok(enriched_changes)
+    // No gold public history yet — return empty instead of falling back to
+    // the legacy `balance_changes` table (kept commented below).
+    Ok(Vec::new())
+
+    //     // Parse dates
+    //     let start_date = params
```
frol, line 179: Is this as designed? If we don't need the code, I'd remove it instead of commenting it out.

frol (commented):
Overall, I feel that our indexing logic is still way too complex. Maybe we should have a call to discuss it with all the knowledge we already accumulated.

nt-be/src/routes/balance_changes.rs
ktarun1419: I left it as-is for now until the entire pipeline is reliable.

We’ll remove this once we’re confident in the public revamp.

Other reason was, If we want to ship a production release earlier, we can comment out the new pipeline, redirect traffic to the old one, and cut a release.

nt-be/src/handlers/balance_changes/history.rs
ktarun1419: We don’t store these as `f64`: the DB USD columns are `NUMERIC`, and the internal Rust models use `BigDecimal`.

The `f64` here is only at the `RecentActivity` API serialization boundary, matching the existing `valueUsd`, chart `priceUsd/valueUsd`, and token metadata price fields. These are display estimates, not accounting values.

frol (approved):

PR #1162, opened 2026-07-21
chore: raise DB slow-statement WARN threshold (1s -> 5s, configurable)

frol-ai:
Addresses the noisy `WARN slow statement` logs from the apalis-board endpoints:
- `GET /api/v1/overview` — ~1.4s
- `GET /api/v1/queues` — ~1.5s

These are the board library's full-table aggregate queries over `apalis.jobs` (`SUM(CASE WHEN status …)` etc.), run only on the admin dashboard. Their SQL lives in the apalis-board crate (not ours to change), and 1.4–1.5s is expected for full-table analytics — but sqlx's default 1s slow-statement threshold flags them on every load, polluting the logs.

## Change
Switch the pool to `connect_with` and set the threshold via `PgConnectOptions::log_slow_statements(Warn, …)`, default **5s** (`DB_SLOW_STATEMENT_THRESHOLD_SECONDS`). Routine multi-second analytics no longer warn; genuinely pathological statements — the seconds-long token-list `DISTINCT…UNION` and the backfill upserts — still warn (they exceed 5s).

The real *speedup* for the board is a smaller `apalis.jobs` table (retention + the stuck-`Running` reclaimer in the companion PR); this change just fixes the log noise.

## Testing
`cargo fmt --check`, `cargo clippy --lib -- -D warnings`, `cargo check`: clean.

Note: builds on the merged pool-hardening change (#1155); same pool builder.

https://claude.ai/code/session_01FRWEKUFYCcmGGhwdDYd471

### DIFF
--- a/nt-be/Cargo.lock
+++ b/nt-be/Cargo.lock
@@ -2879,6 +2879,7 @@ dependencies = [
  "hex",
  "hmac 0.12.1",
  "jsonwebtoken",
+ "log",
  "moka",
  "near-account-id",
  "near-api",
--- a/nt-be/Cargo.toml
+++ b/nt-be/Cargo.toml
@@ -23,6 +23,7 @@ hex = "0.4"
 serde_with = { version = "3.16.1", features = ["base64"] }
 chrono = "0.4"
 futures = "0.3"
+log = "0.4"
 async-stream = "0.3"
 governor = "0.7"
 sqlx = { version = "0.8", features = [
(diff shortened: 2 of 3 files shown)


frol-ai:
Superseded by #1163, which makes the board queries genuinely fast (de-bloating apalis.jobs) instead of raising the slow-statement threshold. Keeping sqlx's 1s threshold is actually preferable — it still flags the genuinely-slow statements (the ~6s silver upsert, ~12s token-list query).

PR #1177, opened 2026-07-22
Fix/charts

ktarun1419:


nt-be/src/main.rs
```diff
@@ -31,6 +31,11 @@ async fn async_main() {
             .expect("Failed to initialize application state"),
     );
 
+    // Every instance serves prices from the in-memory token snapshot; the
+    // ingest cron that writes the underlying table runs only on the jobs
+    // leader, so each process refreshes its own snapshot.
+    state.token_price_service.spawn_snapshot_refresher();
+
```
frol, line 38: Can it be configured as apalis job?

frol: From @ktarun1419:

> If it was an apalis job, only the instance that picked up the job would get fresh data (due to the leadership rules implemented), and the other would serve empty/stale prices (that was also a bug).

frol (approved):

PR #1188, opened 2026-07-23
refactor: History of transactions v3 (confidential + public)

ktarun1419:
Data flow (medallion pipeline)

- Bronze:  raw NearBlocks pages (`nearblocks_ft`, `nearblocks_mt`, `nearblocks_receipt`) per account; cursors track backfill and latest-refresh progress

- Silver, two projections from bronze:
  - `silver_public_transfer_legs` — display-oriented transfer legs that become the activity feed
  - `silver_balance_history` — the balance-complete ledger: every balance-affecting movement (FT/MT deltas, native receipts, wraps, staking, sponsor top-ups), not just feed-visible ones; each entry carries the user-owned user_balance_after and intra_block_seq for exact chain ordering; rebuilt suffix-wise (delete-then-insert from the recompute point)
  
- Verification — compares `silver_balance_history` heads against on-chain balances via RPC; a full-history gate must pass once before charts serve at all; small drifts are rebased into the ledger as reconciliation entries; head drift only marks charts Stale (never revokes the gate)

- Gold — gold_treasury_ledger_events, the single read table:
  - visible activity rows projected from silver legs, stamped with user balances from silver_balance_history (source_kind='public_silver_leg', history_visible=true)
  - Hidden rows for ledger entries with no visible leg (sponsor top-ups, wraps, rebases) so the table alone is balance-complete for charts without double-counting the feed

- Replaced `PUBLIC_HISTORY_MEDALLION_READS` with a single `UNIFIED_GOLD_LEDGER_READS` flag gating public activity, charts, counts, prior balances, proposal enrichment, and asset balances (off by default in render.yaml)

- Chart query pushes window + per-asset carry-forward baseline and token filtering into SQL; readiness now distinguishes "empty requested range" from "no ledger data" (has_gold_balance_points)

- Latest-refresh job queue polls on a fixed 1s interval (no idle backoff); NearBlocks rate ceiling and latest-job concurrency overridable via NEARBLOCKS_MAX_PER_MINUTE / PUBLIC_HISTORY_LATEST_CONCURRENCY

- Repointed kr_analytics_treasury_monthly public flow metrics at the unified ledger (hidden rows excluded)


ktarun1419:
Merge #1250 first before merging this into main

frol (approved):
It is a huge one. I did my best to review it, but it is not realistic to not miss anything, so I trust that it was already well tested. I did not spot any obvious issues, so I believe it is ready for testing on staging

Issue #1199, opened 2026-07-24
[BUG] Trezu - Support Center: Close icon does not display Pointer Cursor


rknearn-qa:
### Summary

The **Close** icon in the **Support Center** does not display a pointer cursor when hovered.

This creates an inconsistent interaction pattern because the icon is clickable, but its cursor behavior does not visually indicate that it is an interactive element. Relates to #915 

### Steps To Reproduce

1. Open the **Support Center**.
2. Locate the **Close** icon.
3. Hover the cursor over the icon.
4. Observe the cursor behavior.

### Actual Result

* The cursor does not change to a **pointer** when hovering over the clickable **Close** icon.

https://github.com/user-attachments/assets/3b382eee-a8fe-41a8-abb6-e99b570986f5

### Expected Result

* The **Close** icon should display a **pointer cursor** on hover.
* Cursor behavior should be consistent with other clickable elements throughout the application.

### Special Notes

```
Possible Root Cause:

The clickable Close icon is missing the CSS `cursor: pointer` style
or the corresponding button/link styling is not applied.
```

### Workarounds
None

### Repro Rate
**Always (10/10)**

### Setup - OS - Browser - Environment

* OS / Devices: Windows 11 (x64)
* Screen resolution(s): 1366 x 768, 1600 x 720, 1792×828
* Browser(s) + version(s): Chrome (150.0.7871.129), Mozilla Firefox (151.0.3), Brave (1.91.168), Edge (149.0.4022.52)
* Wallet: Meteor
* App environment: Staging (trezu.app)
* Near Network: Mainnet

#### User Impact

* Users may not immediately recognize the Close icon as interactive.
* Creates inconsistent cursor behavior compared with other clickable UI elements.

#### System Impact

* Minor UI consistency issue.
* Interactive element styling is inconsistent across the application.

#### Business / Release Risk

* Low risk.
* May slightly reduce UI clarity and perceived polish.

### QA Verdict

**Verdict:** Lowest issue

#### Rationale

Clickable controls should provide consistent visual feedback on hover. The Close icon should use a pointer cursor to clearly communicate that it can be clicked.

### Recommendation

* Add `cursor: pointer` to the clickable Close icon or its interactive wrapper.
* Verify cursor behavior for other Support Center interactive controls.
* Add regression coverage for:

  * Close icon hover state.
  * Other clickable Support Center icons and buttons.
  * Cursor consistency across supported browsers.

### Priority

⚪ P4 : Low

[...]

Issue #1174, opened 2026-07-21
[DEV]  Improve Bulk Payment flow and UI states


olskik12:
Issue 1 
--------
The current UI is different from what we agreed on.

<img width="1470" height="900" alt="Image" src="https://github.com/user-attachments/assets/50ab0866-2d89-4544-9a6c-63e326864828" />

Please add an info icon next to the amount with a tooltip showing the total amount + network fee.

[Figma](https://www.figma.com/design/UlUkb3wL7fQq0xNvmjFnEg/TREZU?node-id=5746-139658&t=JI7TeALGfbk2uCKW-4)

Issue 2
------------
1. When the user opens the Bulk Payment screen, the selected token should be the token that the user has in their balance.
2. The recipient network selector should be disabled by default and become active only after the user adds a recipient address.

<img width="1710" height="930" alt="Image" src="https://github.com/user-attachments/assets/037d63b6-3fc1-4448-9c87-35f8c5ac1aad" />

rknearn-qa:
Implemented

<img width="858" height="636" alt="Image" src="https://github.com/user-attachments/assets/44c6329a-7415-43a5-b0d0-140414c26c53" />
<img width="769" height="634" alt="Image" src="https://github.com/user-attachments/assets/69eb6a15-79ad-4bf5-b46d-94bfc0c2996b" />
<img width="1363" height="689" alt="Image" src="https://github.com/user-attachments/assets/ccdb20c4-f112-45f2-9a92-8368a07ae4ba" />
<img width="1365" height="668" alt="Image" src="https://github.com/user-attachments/assets/aadb652e-3eed-429e-8cc4-a63d609630ee" />
<img width="1253" height="676" alt="Image" src="https://github.com/user-attachments/assets/ca8f1a24-954d-4571-b63d-4738bed00e7e" />

* OS / Devices: Windows 11 (x64), Motorola G23 (Android 14), Iphone XR (iOs 18.7.6)
* Screen resolution(s): 1366 x 768, 1600 x 720, 1792×828
* Browser(s) + version(s): Chrome (150.0.7871.129), Mozilla Firefox (151.0.3), Brave (1.91.168), Safari (18.7.4), Edge (149.0.4022.52)
* Wallet: Meteor
* App environment: Staging (trezu.app)
* Near Network: Mainnet

PR #1201, opened 2026-07-27
Fix/token defaults and bulk payment ux

Megha-Dev-19:
#1199 #1174
- Default token picker (payments, bulk, deposit): highest-USD owned asset from cache, else USDC on NEAR (Intents); skeleton while assets load
- Bulk review: network fee folded into amount with info tooltip instead of a per-recipient fee row
- Bulk recipient network: disabled until a recipient address is added
- Fix modal close icon cursor


frol (approved):

frol:
@Megha-Dev-19 Please, resolve the merge conflict here, and let's get it merged

Issue #1171, opened 2026-07-21
[BUG] fix mobile responsive on bulk payment


olskik12:
Issue 1. When the user selects a token, the modal is too large for the screen size. It should fit the screen properly.
<img width="590" height="1278" alt="Image" src="https://github.com/user-attachments/assets/978b0072-0d85-4b50-a165-4997c4eb1027" />

Issue 2 When the user closes the token selection modal, the whole page layout breaks and becomes too large. It should fit the screen size properly.
<img width="590" height="1278" alt="Image" src="https://github.com/user-attachments/assets/3f3d65ca-4a89-4c57-bc9d-63f53e97f1ae" />

olskik12:
Hi @Megha-Dev-19 
Could you please fix this? When I type something in the input, the page still doesn’t return to the previous state

https://github.com/user-attachments/assets/cdd0719b-964a-4fbc-be9f-ad8893b4245e

rknearn-qa:
Retested
Fixed

<img width="591" height="1280" alt="Image" src="https://github.com/user-attachments/assets/407d0936-8cbb-4626-b01a-e6d6b0fc97f6" />

<img width="591" height="1280" alt="Image" src="https://github.com/user-attachments/assets/b9b98dc7-c657-494a-b1c9-b6cc4847a48d" />

* OS / Devices: Motorola G23 (Android 14), Iphone XR (iOs 18.7.6)
* Screen resolution(s): 1600 x 720, 1792×828
* Browser(s) + version(s): Chrome (150.0.7871.129), Mozilla Firefox (151.0.3), Brave (1.91.168), Safari (18.7.4), Edge (149.0.4022.52)
* Wallet: Meteor
* App environment: Staging (trezu.app)
* Near Network: Mainnet

Issue #1110, opened 2026-07-14
[DEV] Enable Member Invitations via Shareable Links


olskik12:
Implement #936 

What needs to be done:

Members page
--------------------

[Figma](https://www.figma.com/design/UlUkb3wL7fQq0xNvmjFnEg/TREZU?node-id=8990-69661&t=e3ZFnzEgSE8chA9o-4)

* The “Pending” and “Wants To Join” buttons should appear contextually, meaning they should only be visible when there are relevant requests.
* Add a menu for the “Add member” button with two options:
    * Add Manually
    * Invite Member

Add new member
---------------------
[Figma](https://www.figma.com/design/UlUkb3wL7fQq0xNvmjFnEg/TREZU?node-id=9676-116447&t=e3ZFnzEgSE8chA9o-4)

In the new flow, we won't have modals for adding members. Everything will live on separate pages, following the same pattern as all other flows.

1. Add Manually
Everything we currently have stays exactly the same — only moved onto a dedicated page instead of a modal.

2. Invite Member
* A dedicated page with a "How it works" details section.
* The user is able to generate an invite link.
* After the user generates the link, they land on a screen where they can copy the generated link.

Member Wants To Join
----------------------------
[Figma](https://www.figma.com/design/UlUkb3wL7fQq0xNvmjFnEg/TREZU?node-id=9678-116463&t=e3ZFnzEgSE8chA9o-4)
* The user has a "Wants To Join" button on the main page.
* On click, the user sees a form where the addresses are already pre-filled — the user only needs to assign roles and create the request.
* The pre-filled addresses cannot be edited.
* In this case, the user can only add users who have submitted a request; they cannot add users manually.

New member side
---------------------
[Figma](https://www.figma.com/design/UlUkb3wL7fQq0xNvmjFnEg/TREZU?node-id=9361-37159&t=e3ZFnzEgSE8chA9o-4)

* To send a join request, the user must connect their wallet.
* After connecting, the user sees their wallet address, can optionally add a name, and submits the request.
* The user then sees a success state, from which they can navigate to the treasury they received the invite from.
* In that treasury, the user will see a banner stating that they are in a "guest mode" and are waiting to be added.

Corner case:
The invite link is single-use. If someone opens a link that has already been used for a request, they will not have the option to submit a request.

Mobile
-----------
[...]

PR #1207, opened 2026-07-28
feat: Members invite

Megha-Dev-19:
#1110
#1171


nt-fe/app/(treasury)/[treasuryId]/members/edit/page.tsx
```diff
@@ -0,0 +1,293 @@
...
+                members_count: data.members.length,
+            });
+
+            queryClient.invalidateQueries({
+                queryKey: ["proposals", treasuryId],
+            });
+
+            router.push(`/${treasuryId}/members`);
+        } catch (error) {
+            console.error("Failed to edit members:", error);
```
Comment (frol) [inline nt-fe/app/(treasury)/[treasuryId]/members/edit/page.tsx]: Does Sentry catch console.error? If not, then we are basically silencing errors and don't help ourselves to catch errors early.

Also, it is not helpful for the user here either - nothing will happen, not even a toast is shown

nt-fe/app/(treasury)/[treasuryId]/members/edit/page.tsx
```diff
@@ -0,0 +1,293 @@
...
+    );
+
+    return (
+        <PageComponentLayout title={t("title")} description={t("description")}>
+            <div className="max-w-xl mx-auto w-full">
+                <FormProvider {...form}>
+                    {!membersReady || originalMembers.length === 0 ? (
+                        <PageCard>
+                            <div className="h-40 animate-pulse bg-muted rounded-lg" />
+                        </PageCard>
```
Comment (frol) [inline nt-fe/app/(treasury)/[treasuryId]/members/edit/page.tsx]: I don't feel that a loader is helpful to show when the originalMembers.length === 0. We should either return to a previous page or display a message that there are no accepted invites yet or something along those lines.

nt-fe/app/(treasury)/[treasuryId]/members/hooks/use-member-validation.ts
```diff
@@ -164,7 +164,9 @@ export function useMemberValidation(
             // Only check roles that are present in members being removed
             const rolesToCheck = new Set<string>();
             membersToCheck.forEach((member) => {
-                member.roles.forEach((role) => rolesToCheck.add(role));
+                for (const role of member.roles) {
+                    rolesToCheck.add(role);
+                }
```
Comment (frol) [inline nt-fe/app/(treasury)/[treasuryId]/members/hooks/use-member-validation.ts]: Why changing? I also find it a bit weird to mix forEach and for-of loop - stick to one style.

nt-fe/app/(treasury)/[treasuryId]/members/add/page.tsx
```diff
@@ -0,0 +1,378 @@
...
+            });
+
+            queryClient.invalidateQueries({
+                queryKey: ["proposals", treasuryId],
+            });
+
+            router.push(`/${treasuryId}/members`);
+        } catch (error) {
+            console.error("Failed to add members:", error);
+        }
```
Comment (frol) [inline nt-fe/app/(treasury)/[treasuryId]/members/add/page.tsx:314]: same as in another comment

frol (changes requested):
Overall it looks good.

I am still reviewing it, but here are some initial comments that needs to be updated (in addition to the ones raised in the chat)

nt-fe/components/number-badge.tsx
```diff
@@ -31,17 +31,19 @@ export function NumberBadge({
     variant = "default",
     sizes = "default",
     ariaLabel,
+    className,
```
frol: This looks suspicious. I would suggest to ban `className` on all React components outside of `components` folder so that we don't accidentally introduce custom ad-hoc styles, which would lead to inconsistencies in the UI.

Are you sure this is needed and not abused?

nt-fe/app/(treasury)/[treasuryId]/members/join-requests/page.tsx
```diff
@@ -0,0 +1,432 @@
...
+                            policy: updatedPolicy,
+                        },
+                    },
+                },
+                proposalType: "other",
+            });
+        } catch (error) {
+            // Wallet rejected / proposal failed — stay on review to retry.
+            console.error("Failed to create add-members proposal:", error);
+            return;
```
Comment (frol) [inline nt-fe/app/(treasury)/[treasuryId]/members/join-requests/page.tsx:334]: Please, add to the AI skills that all the errors that block user flow and cannot be automatically resolved (e.g. with retry) must be reported to the user.

Also, review all existing console.error / catch-without-error-handling in the frontend and fix them.

P.S. Not every instance of console.error needs to be reported to users (e.g. below `apprveRequests.mutateAsync` is fine to fail without users noticing it - yet we have to make sure we catch it on Sentry)

frol (changes requested):

nt-fe/app/(treasury)/[treasuryId]/members/edit/page.tsx
Comment (Megha-Dev-19) [reply nt-fe/app/(treasury)/[treasuryId]/members/edit/page.tsx]: Updated to use reportError (Sentry + log) and show a toast when the submit fails. Same for add/delete/invite/join.

nt-fe/app/(treasury)/[treasuryId]/members/edit/page.tsx
Comment (Megha-Dev-19) [reply nt-fe/app/(treasury)/[treasuryId]/members/edit/page.tsx]: Skeleton is only shown while policy/members are loading now. If the URL has no resolvable members, we redirect back to /members instead of keeping an empty loader.

nt-fe/components/number-badge.tsx
Megha-Dev-19: Good point, I removed the className. Added a proper shape="pill" variant.

nt-fe/app/(treasury)/[treasuryId]/members/join-requests/page.tsx
Comment (Megha-Dev-19) [reply nt-fe/app/(treasury)/[treasuryId]/members/join-requests/page.tsx:334]: For this PR I fixed the members invite flow end-to-end. I will prefer to do the full frontend pass of every console.error/empty catch in a follow-up, created a ticket for it, hope that's okay. #1213

frol (approved):
The only feedback left is the one we discussed in the chat:

> If the user, who received the invitation link, already has the name - we prepopulate it in the input and disable the input.
> 
> If the user does not yet have the name - we allow them to enter the name in the input.

frol (approved):

PR #1221, opened 2026-07-30
fix(nt-fe): pick up near.com colors, styles, animations and sidebar

denbite:
image


nt-fe/app/wallet/layout.tsx
```diff
@@ -38,7 +38,7 @@ export default async function WalletLayout({
             lang={locale}
             dir={dir}
             suppressHydrationWarning
-            className={`${geistSans.variable} ${geistMono.variable}`}
+            className={`${figtree.variable} ${geistMono.variable}`}
```
frol, line 41: It is suspicious that we use a mix of geistMono and figtree.

frol (approved):
I think it is a pretty good first pass. I want to see it live on the new staging, so I will approve it as is and will iterate it from there.

nt-fe/app/wallet/layout.tsx
denbite: yep, will get it cleaned up in the follow-up PRs

Issue #1236, opened 2026-08-03
[BUG]  Trezu - Invite Member: App crashes when removing a Member from Wants to Join List


rknearn-qa:
### Summary

The **Wants to Join** page crashes when removing a member from the list after roles have been assigned to multiple pending members.

The application throws a client-side `TypeError` while recalculating the state after a member is removed, resulting in a full application error page instead of simply removing the selected member.

### Preconditions

* Treasury has at least **3 members** in the **Wants to Join** list.

### Steps To Reproduce

1. Open **Members → Wants to Join**.
2. Assign a role to each pending member.
3. Click the **Delete** icon for any member.
4. Observe the application.

### Actual Result

* The application crashes.
* The following error page is displayed:

```
Application error: a client-side exception has occurred
(see the browser console for more information).
```

Console error:

```
TypeError: Cannot read properties of undefined (reading 'length')
```

The error occurs during an `Array.every()` operation.

https://github.com/user-attachments/assets/62a264bf-55b7-40a0-8c1b-855715ef6780

<img width="1365" height="700" alt="Image" src="https://github.com/user-attachments/assets/e684607a-c2b5-4eef-b744-693724a1f1ba" />

### Expected Result

* The selected member should be removed from the **Wants to Join** list.
* The remaining members and their assigned roles should stay unchanged.
* The application must not crash when the list is modified.

### Special Notes

```text
Possible Root Cause:

After removing a member, the component recalculates the role/member
validation state.

One of the values expected by the validation logic becomes undefined,
but the code still attempts to access `.length` on it.

The stack trace shows the failure occurring inside an Array.every()
operation, suggesting that the validation logic assumes every pending
member/role entry contains a defined collection.

The state should be normalized after member removal and validation
should safely handle undefined/empty role collections.
```

### Workarounds

* Reload the page after the crash.
* Avoid removing members after assigning roles if possible.

### Repro Rate

**Always (10/10)** with 3+ pending members after assigning roles.

### Setup - OS - Browser - Environment

* OS / Devices: Windows 11 (x64), Motorola G23 (Android 14), Iphone XR (iOs 18.7.6)
* Screen resolution(s): 1366 x 768, 1600 x 720, 1792×828
* Browser(s) + version(s): Chrome (150.0.7871.129), Mozilla Firefox (151.0.3), Brave (1.91.168), Safari (18.7.4), Edge (149.0.4022.52)
* Wallet: Meteor
* App environment: Staging (trezu.app)
[...]

Issue #1235, opened 2026-08-03
[BUG] Trezu - Invite Member: Existing pending Member can submit another Join Request via New Invitation Link


rknearn-qa:
### Summary

A user who already has a **pending join request** for a Treasury can use a newly generated invitation link for the same Treasury and attempt to submit another join request.

The application detects that a pending request already exists, but the user is still allowed to open and submit the Join flow again instead of being prevented from creating another request.

### Preconditions

* Treasury exists.
* Invitation link has been created.
* Potential member has access to the invitation link.
* The potential member does not yet have an approved membership.

### Steps To Reproduce

1. Open the invitation link as a potential member.
2. Click **Ask to Join**.
3. Wait until the join request is created and remains **Pending**.
4. As a Treasury member, generate a **new invitation link** for the same Treasury.
5. Open the new invitation link using the same potential member account.
6. Attempt to submit **Ask to Join** again.

### Actual Result

* The user can open the new invitation flow.

* After attempting to join, the application shows:

  **A pending join request already exists for this treasury**

* The user is still allowed to enter and attempt the Join flow despite already having a pending request.

<img width="1365" height="691" alt="Image" src="https://github.com/user-attachments/assets/c04cc028-1262-41e5-b840-aea106a76ee6" />

https://github.com/user-attachments/assets/07891dce-09d5-480e-a875-ab5a38f17623

### Expected Result

* A user with an existing pending join request should not be able to start another Join flow for the same Treasury.
* Opening a new invitation link for the same Treasury should immediately recognize the existing pending request.
* The user should be shown the **Pending** membership state instead of being presented with another **Ask to Join** flow.

### Special Notes

```text
Possible Root Cause:

The invitation link appears to act as an independent entry point and does
not prevent navigation into the Join flow when the same account already
has a pending request for the Treasury.

The existing pending membership state should be checked before rendering
or enabling the "Ask to Join" action.

Expected state resolution:

Invitation link opened
        ↓
Check membership status
        ↓
Already a member? → Member state
Pending join request? → Pending state
No membership/request? → Ask to Join
```

Generating a new invitation link should not reset or bypass the existing membership-request state.

### Workarounds

* Wait for the existing join request to be approved or rejected before attempting to join again.

### Repro Rate

**Always (10/10)**

### Setup - OS - Browser - Environment

* OS / Devices: Windows 11 (x64), Motorola G23 (Android 14), Iphone XR (iOs 18.7.6)
[...]

Issue #1234, opened 2026-08-03
[BUG] Trezu - Add Member: Address placeholder incorrectly contains `.near`


rknearn-qa:
### Summary

The **Add Member** form displays an incorrect placeholder for the member address input.

The placeholder includes the `.near` suffix (`address.near`), although the expected placeholder should contain only `address`.

### Preconditions

* User has permission to manage treasury members (Governance Role).

### Steps To Reproduce

1. Open the treasury.
2. Go to **Members → Add New Member**.
3. Observe the **Address** input field.

### Actual Result

* Address input placeholder displays:

```
address.near
```

<img width="1365" height="674" alt="Image" src="https://github.com/user-attachments/assets/94ee127d-c584-47cd-baf0-71e75e90e8f6" />

### Expected Result

* Address input placeholder should display:

```
address
```

* The placeholder should describe the expected input without including a specific network/domain suffix.

<img width="904" height="463" alt="Image" src="https://github.com/user-attachments/assets/4f0320e4-b024-43f9-a7ac-9eb8c78be364" />

[Design Link](https://www.figma.com/design/UlUkb3wL7fQq0xNvmjFnEg/TREZU?node-id=8991-84628&t=qI1Icr9Wt9ajcQjM-4)

### Special Notes

```
UI/content issue.

The placeholder text appears to use "address.near" instead of the
expected generic "address".
```

### Workarounds

None.

### Repro Rate

**Always (10/10)**

### Setup - OS - Browser - Environment

* OS / Devices: Windows 11 (x64), Motorola G23 (Android 14), Iphone XR (iOs 18.7.6)
* Screen resolution(s): 1366 x 768, 1600 x 720, 1792×828
* Browser(s) + version(s): Chrome (150.0.7871.129), Mozilla Firefox (151.0.3), Brave (1.91.168), Safari (18.7.4), Edge (149.0.4022.52)
* Wallet: Meteor
* App environment: Staging (trezu.app)
* Near Network: Mainnet

#### User Impact

* Minor UI inconsistency.
* May imply that only `.near` addresses are expected.

#### System Impact

* No functional impact.

#### Business / Release Risk

* Low — cosmetic/content issue only.

### QA Verdict

**Verdict:** ❌ Needs Fix

#### Rationale

The placeholder should accurately communicate the expected input format and remain generic rather than implying a specific address suffix.

### Recommendation

* Change the placeholder from `address.near` to `address`.
[...]

Issue #1233, opened 2026-08-03
[BUG] Trezu - Invite Member: Missing Loading Skeleton during Invite Link Generation


rknearn-qa:
### Summary

When generating a new invitation link, the **Your Invite Link** field immediately changes to the newly generated link without displaying any loading state.

On mobile devices, the invitation link is cropped, so the change is not visually obvious. This can make the user think that nothing happened after tapping **Generate New Link**.

According to the approved design, the input should display a **loading skeleton** while the new invitation link is being generated.

### Preconditions

* User has at least the **Governance** role.
* User can generate invitation links.
* Invite Member page is accessible.

### Steps To Reproduce

1. Open the **Invite Member** page.

   ```
   https://testenv.trezu.app/bulk-payment2.sputnik-dao.near/members/invite
   ```

2. Tap **Generate New Link**.

3. Immediately observe the **Your Invite Link** input while the new link is being generated.

### Actual Result

* No loading state or skeleton is displayed.
* The input changes directly to the newly generated link.
* On mobile, the link is cropped, making the change difficult to notice.
* There is no clear visual indication that link generation is in progress.

<img width="1365" height="676" alt="Image" src="https://github.com/user-attachments/assets/eaf55895-464f-42a4-9b75-7b5effc66daa" />

https://github.com/user-attachments/assets/d810e823-23b5-4d4c-a9c5-41b9d61a4ec6

### Expected Result

* The **Your Invite Link** field should display a loading **skeleton** while the new invitation link is being generated.
* The previous link should not appear as if it is still active during the generation process.
* Once generation completes, the skeleton should be replaced with the newly generated invitation link.

<img width="1044" height="550" alt="Image" src="https://github.com/user-attachments/assets/04cb09e5-b50e-4920-9acb-1b46f8ccc020" />

[Design Link](https://www.figma.com/design/UlUkb3wL7fQq0xNvmjFnEg/TREZU?node-id=9702-24137&t=qI1Icr9Wt9ajcQjM-4)

### Special Notes

```
Possible Root Cause:

The invitation-link generation request does not expose its pending state
to the input component, or the component does not render a loading state
while the request is in progress.

The approved design expects the input to provide explicit visual feedback
during link generation.
```

### Workarounds

None.

### Repro Rate

**Always (10/10)**

### Setup - OS - Browser - Environment

* OS / Devices: Windows 11 (x64), Motorola G23 (Android 14), Iphone XR (iOs 18.7.6)
[...]

PR #1240, opened 2026-08-03
fix: resolve invite member QA bugs

Megha-Dev-19:
- Fix Wants to Join crash when removing a member after roles are set (#1236)
- Keep member count stable during Create Request (#1230)
- Show Pending / already-member states for invite joiners instead of re-opening Ask Join (#1235, #1232)
- Add invite-link generation skeleton and constrain iOS Safari selection overflow (#1233, #1231)
- Change Add Member address placeholder from `address.near` to `address` (#1234)

### DIFF
--- a/nt-fe/messages/ko.json
+++ b/nt-fe/messages/ko.json
@@ -843,7 +843,8 @@
             "yourLink": "초대 링크",
             "copyLink": "링크 복사",
             "linkCopied": "초대 링크가 복사되었습니다",
-            "generateNewLink": "새 링크 생성"
+            "generateNewLink": "새 링크 생성",
+            "generatingNewLink": "Generating New Link"
         },
         "joinRequests": {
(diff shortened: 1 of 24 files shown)


nt-fe/messages/ko.json
```diff
@@ -843,7 +843,8 @@
             "yourLink": "초대 링크",
             "copyLink": "링크 복사",
             "linkCopied": "초대 링크가 복사되었습니다",
-            "generateNewLink": "새 링크 생성"
+            "generateNewLink": "새 링크 생성",
+            "generatingNewLink": "Generating New Link"
```
frol: @Megha-Dev-19 Translations are not applied for the `generatingNewLink` key. Update it for all the locales.

frol (approved):
Once the translations are fixed, feel free to merge it

PR #1241, opened 2026-08-03
ci: fix failing maxmind update script by bumping Node.js to 24.19.0

denbite:
The frontend build's `node ./node_modules/geoip-lite/scripts/updatedb.js` step crashed with `ERR_USE_AFTER_CLOSE` while iterating the MaxMind CSVs. Cause is a Node `readline` async-iterator bug present in 24.2.0–24.14.1 ([nodejs/node#60507](https://github.com/nodejs/node/issues/60507)), fixed in 24.15.0 ([#60548](https://github.com/nodejs/node/pull/60548)); nothing pinned Node, so Render resolved 24.14.1. Pins Node 24.19.0 via `NODE_VERSION` on the three `runtime: node` services in `render.yaml` and adds a matching `nt-fe/.node-version`.


frol:
It somehow got fixed on its own, so let's not touch it for now. I will keep it as a draft for now

frol (approved):
The issue got back to us and I tested that this fix works, so let's merge it

Issue #1210, opened 2026-07-29
[DEV] Implement new deposit flow


olskik12:
Implement #998 
[Prototype for confidential treasury](https://www.figma.com/proto/UlUkb3wL7fQq0xNvmjFnEg/TREZU?node-id=10080-15054&viewport=-4118%2C-6446%2C0.22&t=TkW8AG8ErV6XRa2e-8&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=10080%3A15054&show-proto-sidebar=1&page-id=1056%3A60394&hide-ui=1)
Figma - [For confidential treasury](https://www.figma.com/design/UlUkb3wL7fQq0xNvmjFnEg/TREZU?node-id=10080-14990&t=hA0VAzUPYwztuTrt-4), [public](https://www.figma.com/design/UlUkb3wL7fQq0xNvmjFnEg/TREZU?node-id=7634-65593&t=hA0VAzUPYwztuTrt-4), [Mobile ](https://www.figma.com/design/UlUkb3wL7fQq0xNvmjFnEg/TREZU?node-id=9647-43370&t=hA0VAzUPYwztuTrt-4), [Guest mode](https://www.figma.com/design/UlUkb3wL7fQq0xNvmjFnEg/TREZU?node-id=10133-18796&t=hA0VAzUPYwztuTrt-4), [Transfer Link](https://www.figma.com/design/UlUkb3wL7fQq0xNvmjFnEg/TREZU?node-id=10133-55280&t=hA0VAzUPYwztuTrt-4), [Loading State](https://www.figma.com/design/UlUkb3wL7fQq0xNvmjFnEg/TREZU?node-id=9647-44572&t=hA0VAzUPYwztuTrt-4)

We need to implement the new version of the deposit flow.

* In a confidential treasury, users should have tabs to choose how they want to make a deposit: from a confidential wallet or from a public wallet.
* Users should see alerts and confirm that they understand the deposit conditions before proceeding.
* After a user creates a deposit address, they should be able to open the link with the deposit conditions or copy the deposit address.

In a public treasury, users will simply create a deposit address and have the option to copy or share the address.

olskik12:
@Megha-Dev-19 
1. Also, together with this, we should update the token and network names to match the formatting used on near.com. Right now, all of our token and network names are in ALL CAPS, which makes them harder to read. On near.com, not all names are capitalised, making them much easier to scan and read.
2. #798 Can we also do this ticket as part of this work?
[...]

Issue #1129, opened 2026-07-17
[BUG] Trezu - Confidential Treasury: Guest Dashboard does not display Pending Request count


rknearn-qa:
### Summary

When a guest (non-member) opens a **Confidential Treasury**, the **Pending Requests** section on the Dashboard is collapsed without indicating how many requests are currently pending.

According to the approved design, guests should still be able to see the number of pending requests, even though they cannot view their details.

### Preconditions

* Confidential Treasury exists.
* Treasury contains one or more **Pending Requests**.
* User is **not** a member of the treasury.

### Steps To Reproduce

1. Open a **Confidential Treasury** Dashboard.
2. Ensure the user is **not signed in as a member**.
3. Observe the **Pending Requests** section.

### Actual Result

* The Pending Requests section is displayed in its collapsed/locked guest state.
* The number of pending requests is not shown.

<img width="1365" height="688" alt="Image" src="https://github.com/user-attachments/assets/3782930f-0898-45e1-b39f-5d2876f9d0aa" />

### Expected Result

* The guest version of the **Pending Requests** section should display the total number of pending requests.
* Request details should remain hidden.
* The UI should match the approved design.

<img width="1764" height="1110" alt="Image" src="https://github.com/user-attachments/assets/1b2a42e1-336a-4b4a-9add-f50a4ee97718" />

### Special Notes

```text
Possible Root Cause:

The guest version of the Pending Requests component does not retrieve
or display the pending request count.

The component likely hides the entire dataset instead of exposing only
the aggregated count intended for guest users.
```

### Design Reference

Figma:

```
https://www.figma.com/design/UlUkb3wL7fQq0xNvmjFnEg/TREZU?node-id=5392-163184&t=AJYojOzpRczPjqSD-4
```

### Workarounds
None.

### Repro Rate
**Always (10/10)**

### Setup - OS - Browser - Environment

* **OS / Devices:** Windows 11 (x64), Motorola G23 (Android 14), iPhone XR (iOS 18.7.6)
* **Screen resolution(s):** 1366 × 768, 1600 × 720, 1792 × 828
* **Browser(s) + version(s):** Chrome (149.0.7827.54), Mozilla Firefox (151.0.3), Brave (1.91.168), Safari (18.7.4), Edge (149.0.4022.52)
* **Wallet:** Guest (not a treasury member)
* **App Environment:** Staging (trezu.app)
* **Near Network:** Mainnet

#### User Impact

[...]

Issue #1049, opened 2026-07-06
[DEV] Restore the previous UI for Guest Mode users in the confidential treasury


olskik12:
Problem
------------
Due to a misunderstanding in ticket #684 , the wrong UI was implemented for users in Guest Mode. As a result, all data was hidden.

<img width="1676" height="906" alt="Image" src="https://github.com/user-attachments/assets/2e448a20-d9ff-4620-b4ea-5480d365f157" />

What needs to be done
-------------------------

- [ ] Restore the previous UI. We should only hide the amount and the recipient. All other information should remain visible, as it is public.

[Figma](https://www.figma.com/design/UlUkb3wL7fQq0xNvmjFnEg/TREZU?node-id=5392-161424&t=CwSQWsvuWb37x9oE-4)

<img width="600" height="376" alt="Image" src="https://github.com/user-attachments/assets/c410855e-655c-4ce3-9b76-67af6b822ad9" />

Additional note
------------------

- [ ] In the filters, the token and the recipient should also remain hidden.
- [ ] As part of this task, please also remove the small shield icons with tooltips from the Dashboard, Payment & Exchange  pages (see screenshots). They are no longer needed.

<img width="1672" height="908" alt="Image" src="https://github.com/user-attachments/assets/20d051f4-aa51-4cf1-b4e3-a8be4ee74710" />

<img width="1679" height="821" alt="Image" src="https://github.com/user-attachments/assets/9e0b02c4-9b05-49e0-ad9b-f87b2cf2b270" />
<img width="1676" height="882" alt="Image" src="https://github.com/user-attachments/assets/bc3bbbcf-a02c-4fdd-a3df-468438898d0e" />

olskik12:
@ktarun1419  On the dashboard page on the Pending Requests section we need to show how namy requests the treasury have with pending status, if a request is confidential we use confidential pattern

<img width="1665" height="925" alt="Image" src="https://github.com/user-attachments/assets/465c8e8d-f068-4014-affe-1afef43e4e60" />

[Figma](https://www.figma.com/design/UlUkb3wL7fQq0xNvmjFnEg/TREZU?node-id=5392-163184&t=AJYojOzpRczPjqSD-4)

<img width="882" height="555" alt="Image" src="https://github.com/user-attachments/assets/f363114f-48da-4ec7-8c8d-be2310b10c6a" />

rknearn-qa:
Blocked by #1129

rknearn-qa:
[...]

PR #1243, opened 2026-08-03
feat: redesign deposit flow with source selection

Megha-Dev-19:
- Redesigned the deposit flow into a stepped public / confidential / guest experience, adds a shareable transfer page with Pay with Trezu, and makes one-time confidential deposit addresses valid for 14 days. #1210
- Removed `includeNearNetwork`, which previously added a NEAR network option for tokens that don’t natively return NEAR under them (as shown in the confidential tab).
This now breaks the flow because the confidential tab is top-level, not nested under network selection.
- Fixed confidential display of pending requests in dashboard and added count badge #1049 #1129
- Fixed failing e2e report test


frol (approved):
It is a huge effort! It looks good on the surface, but it is not feasible to review it extensively, so I will test it more on staging.

PR #1250, opened 2026-08-04
added:gated confidential migration to unified table

ktarun1419:


frol (approved):
I have hard time to follow, but I really hope that the codebase will become a lot more cleaner after we completely remove the legacy systems

